PR #10, opened 2025-01-30
Draft: initial row-base pagination concept commit for feedback

jschneidereit:


src/main/java/com/salesforce/datacloud/jdbc/core/fsm/README.md
```diff
@@ -0,0 +1,23 @@
+```mermaid
+---
+config:
+  look: handDrawn
+---
+stateDiagram
```
jschneidereit, line 6: This is out of date and doesn't match State.java, but I'd like it to once this PR is out of "Draft"

src/main/java/com/salesforce/datacloud/jdbc/core/listener/AdaptiveQueryStatusListener.java
```diff
@@ -44,6 +44,7 @@
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 
+@Deprecated
```
jschneidereit, line 47: Good opportunity to go from `0.x` to `1.x` by removing these QueryStatusListeners and the entire QueryStatusPoller concept

src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -87,6 +88,10 @@ public boolean isReady() {
         return listener.isReady();
     }
 
+    private boolean useFSM() {
+        return this.dataCloudConnection.getProperties().containsKey("fsm");
```
jschneidereit, line 92: Will add this to all of the other tests that exercise actual queries later today

src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudQueryStatus.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Optional;
+import lombok.Value;
+import lombok.val;
+
+@Unstable
+public interface DataCloudQueryStatus {
+    String getQueryId();
+
+    long getChunkCount();
+
```
mkaufmann, line 30: Why do we need an interface for a value class?

src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudResultSet.java
```diff
@@ -20,7 +20,7 @@
 public interface DataCloudResultSet extends ResultSet {
     String getQueryId();
 
-    String getStatus();
+    String getCompletionStatus();
```
mkaufmann, line 23: Can we just return an enum here?

src/main/java/com/salesforce/datacloud/jdbc/core/PartialResultSet.java
```diff
@@ -0,0 +1,105 @@
...
+        try {
+            val channel = ExecuteQueryResponseChannel.of(StreamUtilities.toStream(iterator));
+            val reader = new ArrowStreamReader(channel, new RootAllocator(ROOT_ALLOCATOR_MB_FROM_V2));
+            val schemaRoot = reader.getVectorSchemaRoot();
+            val columns = ArrowUtils.toColumnMetaData(schemaRoot.getSchema().getFields());
+            val timezone = TimeZone.getDefault();
+            val state = new QueryState();
+            val signature = new Meta.Signature(
+                    columns, null, Collections.emptyList(), Collections.emptyMap(), null, Meta.StatementType.SELECT);
+            val metadata = new AvaticaResultSetMetaData(null, null, signature);
```
mkaufmann, line 93: Can we factor this logic out to be sharable between the result set implementations?

src/test/java/com/salesforce/datacloud/jdbc/core/listener/QueryStatusListenerTest.java
```diff
@@ -16,6 +16,7 @@
 package com.salesforce.datacloud.jdbc.core.listener;
 
 import com.salesforce.datacloud.jdbc.core.HyperGrpcTestBase;
+import com.salesforce.datacloud.jdbc.core.fsm.QueryStateMachine;
```
mkaufmann, line 19: Please add following test scenarios
- fetch row based where less than limit are returned
- fetch row based where limit can be honored
- fetch row based where multiple rpc calls are made to honor limit

src/main/java/com/salesforce/datacloud/jdbc/core/fsm/RowBased.java
```diff
@@ -0,0 +1,184 @@
...
+    private static class Poll implements RowBasedState {
+
+        @Override
+        public RowBasedState transition(RowBasedContext context) {
+            val statusStream = StreamUtilities.toStream(context.getClient().getQueryInfo(context.getQueryId()))
+                    .map(DataCloudQueryStatus::of)
+                    .filter(Optional::isPresent)
+                    .map(Optional::get);
+            val status = StreamUtilities.takeWhile(statusStream, s -> !isReady(s, context))
+                    .findFirst();
```
mkaufmann, line 163: We shouldn't do status calls if the user explicitly provided a range. Instead we just take the users word that the range is fully available

src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -159,6 +166,16 @@ public static DataCloudConnection of(String url, Properties properties) throws S
                 .build();
     }
 
+    protected DataCloudResultSet getRowBasedResultSet(String queryId, long offset, long limit, boolean fullRange) {
```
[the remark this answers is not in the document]
jschneidereit: Yeah I agree an enum is better, I'm having a hard time deciding which package to put one in though

src/main/java/com/salesforce/datacloud/jdbc/core/PartialResultSet.java
jschneidereit: Yes, absolutely!

jschneidereit: Actually, I'm marking the old implementation that has this shared logic as deprecated so I"ll skip a nice abstraction here

PR #13, opened 2025-02-05
Rename pattern from `serverSetting.` to `querySetting` for consistency

mkaufmann:
Server Settings are easy to be confused with global system-wide settings while users only have the mean to rename connection settings. In V3 Connect API the concept of query level settings is called "querySettings", thus we propose the same name in JDBC. We still accept the old name for transition purposes but will log a warning message.

Drive-By:
 - Factor our the test Hyper Server Process management into separate class for clarity
 - Improve shutdown speed by using waitFor instead of fixed 3s sleep
 - Be strict on config settings as previously we had a silently a setting that hyper wouldn't recognize
 - Use auto port picking for the server process to avoid accidental colissions
 - Be consciously less robust on server process management during test to make any failures and problems very explicitly

### DIFF
--- a/src/test/resources/simplelogger.properties
+++ b/src/test/resources/simplelogger.properties
@@ -1,4 +1,4 @@
 org.slf4j.simpleLogger.logFile=System.out
-org.slf4j.simpleLogger.defaultLogLevel=error
+org.slf4j.simpleLogger.defaultLogLevel=info
 
-org.slf4j.simpleLogger.log.com.salesforce.cdp.queryservice.core.QueryServiceConnection=debug
+org.slf4j.simpleLogger.log.com.salesforce.cdp.queryservice.core.QueryServiceConnection=info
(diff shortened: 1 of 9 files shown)


src/test/resources/simplelogger.properties
```diff
@@ -1,4 +1,4 @@
 org.slf4j.simpleLogger.logFile=System.out
-org.slf4j.simpleLogger.defaultLogLevel=error
+org.slf4j.simpleLogger.defaultLogLevel=info
```
KaviarasuSakthivadivel, line 2: Previously, partners using the driver reported that too many logs were being printed. As a result, we've adjusted it to log only errors. We’ll need to decide if we want to keep it this way.

jschneidereit: This is just for test, though I do feel that way sometimes when looking at the CI, perhaps we can make this set to `error` in CI only? I don't think we should block the PR on that though.

jschneidereit (dismissed):

src/test/resources/simplelogger.properties
mkaufmann: For testing I've now changed it to `warn`. Agree that info is likely too verbose, but I think `warn` is better than `error` because we should strive to only have minimal warnings

jschneidereit (approved):

PR #15, opened 2025-02-06
fix: Truncate query in exceptions and add test coverage for protocol limits

mkaufmann:
With this change we are checking limits around:
- headers
- request size
- parameter sizes

The driver already passed all limits but the tests uncoverd a significant inefficiency around exception creation. With multi megabyte queries this was causing multi second hangs. Thus we now truncate the query in the exception message.
If the query is truncated it'll have a `<truncated>` suffix

### DIFF
--- a/src/main/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandler.java
+++ b/src/main/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandler.java
@@ -28,6 +28,17 @@ import salesforce.cdp.hyperdb.v1.ErrorInfo;
 @Slf4j
 @UtilityClass
 public class QueryExceptionHandler {
+    // We introduce a limit to avoid truncating important details from the log due to large queries.
+    // When testing with 60 MB queries the exception formatting also took multi second hangs.
+    private static final int MAX_QUERY_LENGTH_IN_EXCEPTION = 16 * 1024;
+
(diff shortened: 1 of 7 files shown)


src/main/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandler.java
```diff
@@ -28,6 +28,17 @@
 @Slf4j
 @UtilityClass
 public class QueryExceptionHandler {
+    // We introduce a limit to avoid truncating important details from the log due to large queries.
+    // When testing with 60 MB queries the exception formatting also took multi second hangs.
+    private static final int MAX_QUERY_LENGTH_IN_EXCEPTION = 16 * 1024;
+
+    public static DataCloudJDBCException createQueryException(String query, Exception e) {
+        String exceptionQuery = query;
```
KaviarasuSakthivadivel, line 36: I think we can simplify this here? 

```
String exceptionQuery = query.length() > MAX_QUERY_LENGTH_IN_EXCEPTION 
                        ? query.substring(0, MAX_QUERY_LENGTH_IN_EXCEPTION) + "<truncated>" 
                        : query;

```

KaviarasuSakthivadivel (approved):

KaviarasuSakthivadivel (approved):

PR #18, opened 2025-02-12
feat: add row-based access of results to connection

jschneidereit:
This pull request includes new methods for acquiring query status and "row-based" result sets.


jschneidereit:
Closes #10

src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -173,6 +175,24 @@ private DataCloudPreparedStatement getQueryPreparedStatement(String sql) {
         return new DataCloudPreparedStatement(this, sql, new DefaultParameterManager());
     }
 
+    /**
+     * Use getQueryStatus to determine if your query is "ready" then use this to get a collection of rows.
+     * When using {@link RowBased.Mode#FULL_RANGE} this method is not responsible for calculating the offset near the end of available rows,
+     * you must calculate the correct "pages" of offset and limit.
```
vogelsgesang, line 181: I don't understand what this part of the comment is supposed to tell me. And if I don't get it, I don't think our users will understand it, either

src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudQueryStatus.java
```diff
@@ -0,0 +1,72 @@
...
+
+    long chunkCount;
+
+    long rowCount;
+
+    double progress;
+
+    CompletionStatus completionStatus;
+
+    public boolean isResultsProduced() {
```
vogelsgesang, line 42: `isResultProduced` or `areResultsProduced` but not `isResultsProduced`

vogelsgesang, line 42: we should document the meaning of "isResultsProduced" and "isExecutionFinished" using JavaDoc comments. This class will now be part of our public interface, and the meaning is not obvious

KaviarasuSakthivadivel: changed it to `isResultProduced`

KaviarasuSakthivadivel: I have documented it by looking at the comments from proto files, please let me know if that's okay. @vogelsgesang

src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
KaviarasuSakthivadivel: I have changed this comment as well. Please take a look. @vogelsgesang

KaviarasuSakthivadivel (dismissed):

PR #27, opened 2025-03-11
Adjust `fromChannel` method to minimize injected headers and add examples

mkaufmann:
The intent of this change is to analyze how to use the JDBC driver in the least invasive way (compared to manual gRPC calling). To reach that minimum the `DataCloudConnection.fromChannel` method was adjusted to only add the must have interceptors, which - without usage of the corresponding connection parameters - are complete no ops.

To demonstrate this flow a new `examples` test package was create that demonstrates the approach and serves as documentation. To write that test case (and minimize the use of helpers which would hide complexity) the test hyper instance variable with its getPort method was exposed as public.

Drive-By:t
- The hyper test instance was made non-static to avoid problems in parallel test execution.
- A sample implementation of using the row based API to serve row based paginated results was also added and commented. With the parallel ongoing work we should see simplification of it.

### DIFF
--- a/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
+++ b/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
@@ -150,9 +150,9 @@ public class DataCloudConnection implements Connection, AutoCloseable {
     static List<ClientInterceptor> getClientInterceptors(
             AuthorizationHeaderInterceptor authInterceptor, Properties properties) {
         val list = getPropertyDerivedClientInterceptors(properties);
-        list.add(TracingHeadersInterceptor.of());
+        list.add(0, TracingHeadersInterceptor.of());
         if (authInterceptor != null) {
-            list.add(authInterceptor);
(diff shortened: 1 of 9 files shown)


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -122,19 +127,38 @@ public static DataCloudConnection fromTokenSupplier(
...
     }
 
-    static List<ClientInterceptor> getClientInterceptors(
-            AuthorizationHeaderInterceptor authInterceptor, Properties properties) {
+    /**
+     * Initializes a list of interceptors that handle channel level concerns that can be defined through properties
+     * @param properties - The connection properties
+     * @return a list of client interceptors
+     */
+    static List<ClientInterceptor> getPropertiesClientInterceptors(Properties properties) {
```
jschneidereit, line 135: Just a reminder to consider renaming to `getPropertyDerivedClientInterceptors`

jschneidereit (approved):

PR #28, opened 2025-03-12
revamp query status polling to take advantage of getQueryInfo streaming

jschneidereit:
This pull request introduces a Connection based query status polling capability and makes an attempt to refactor the outmoded  `QueryStatusPoller` concept


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -203,6 +205,33 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId, l
...
+    /**
+     * Checks if all the query's results are ready, the row count and chunk count will be stable.
+     * @param queryId The identifier of the query to check
+     * @param offset The starting row offset.
+     * @param limit The quantity of rows relative to the offset to wait for
+     * @param timeout The duration to wait for the engine have results produced.
+     * @param allowLessThan Whether or not to return early when the available rows is less than {@code offset + limit}
+     * @return The final {@link DataCloudQueryStatus} the server replied with.
+     */
+    @Unstable
```
mkaufmann, line 217: What would be missing to remove the unstable marker here?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/client/DataCloudQueryStatus.java
```diff
@@ -31,6 +32,7 @@
  * </ul>
  */
 @Value
+@Unstable
```
mkaufmann, line 35: What's missing to mark this as stable? I just did a spot check and this looks good, would support marking it as stable

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudPreparedStatement.java
```diff
@@ -121,19 +120,19 @@ protected HyperGrpcClientExecutor getQueryExecutor() {
...
     public ResultSet executeQuery() throws SQLException {
         val client = getQueryExecutor();
-        val timeout = Duration.ofSeconds(getQueryTimeout());
 
         val useSync = optional(this.dataCloudConnection.getProperties(), Constants.FORCE_SYNC)
                 .map(Boolean::parseBoolean)
                 .orElse(false);
-        resultSet = useSync ? executeSyncQuery(sql, client) : executeAdaptiveQuery(sql, client, timeout);
+        resultSet =
+                useSync ? executeSyncQuery(sql, client) : executeAdaptiveQuery(sql, client, getQueryTimeoutDuration());
```
mkaufmann, line 135: Please create a work item so that we revisit the topic of timeouts very soon. I think long term this should be getNetworkTimeout() - but that needs adoption of the `query_timeout` session setting that will soon be released to rpod

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/partial/DataCloudQueryPolling.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+
+    @SneakyThrows
+    static DataCloudQueryStatus waitForQueryStatusWithoutRetry(
+            HyperServiceGrpc.HyperServiceBlockingStub stub,
+            String queryId,
+            Instant deadline,
+            AtomicReference<DataCloudQueryStatus> last,
+            AtomicInteger times,
```
mkaufmann, line 124: Why do these need to be atomic? This looks like non parallel code to me

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/partial/DataCloudQueryPolling.java
```diff
@@ -0,0 +1,140 @@
...
+            boolean allowLessThan)
+            throws DataCloudJDBCException {
+        return waitForQueryStatus(stub, queryId, timeout, status -> {
+            if (allowLessThan) {
+                return status.getRowCount() > offset;
+            } else {
+                return status.getRowCount() >= offset + limit;
+            }
+        });
+    }
```
mkaufmann, line 56: This looks like it doesn't handle the case correctly where we run into a timeout before the query completes (i.e. we exit due to `while (Instant.now().isBefore(deadline)) {` being false.

In that case I would expect it to throw an error that either "no new rows are available" (allowLessThan mode) or "not enough rows are available (for allowLessThan = false mode)

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/hyper/HyperTestBase.java
```diff
@@ -112,13 +118,34 @@ public String getToken() {
         }
     }
 
-    @SneakyThrows
+    private final List<HyperServerProcess> servers = new ArrayList<>();
```
mkaufmann, line 121: Why would we need multiple servers?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/partial/DataCloudQueryPolling.java
jschneidereit: It's because they're used in lambdas, the error you'd get if you used the raw type would be:
>Variable used in lambda expression should be final or effectively final

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/hyper/HyperTestBase.java
jschneidereit: This is so each unit test can create its own instance with its own mock wired up to that instance, perhaps with different settings.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/partial/DataCloudQueryPolling.java
mkaufmann: Ah ok, I think I would prefer to first assign to an effectively final local variable from the lambda and then after the stream change the member. That way the stream complexity workarounds is not polluting the broader code area and can be contained to the area that is using the stream

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/hyper/HyperTestBase.java
mkaufmann: Hmm ok, we can keep that here, but as an alternative I think if you need a unit test specific instance away we should just have a `HyperTestInstance` object that one can create in the unit test and that would then have the convenience methods to create connection / stmt.

I think that would turn into a slightly large refactoring so I'm fine with keeping this here and can do the refactor as a follw up.

Feel free to resolve

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/partial/DataCloudQueryPolling.java
jschneidereit: Fixed!

afurdell (approved):

afurdell (approved):

PR #29, opened 2025-03-13
Update ReadMe for recent update to JDBC driver.

draliatp:


README.md
```diff
@@ -1,12 +1,12 @@
-# Salesforce DataCloud JDBC Driver
+# Salesforce Data Cloud JDBC Driver
 
-With the Salesforce Data Cloud JDBC driver you can efficiently query millions of rows of data with low latency, and perform bulk data extractions.
-This driver is read-only, forward-only, and requires Java 8 or greater. It uses the new [Data Cloud Query API SQL syntax](https://developer.salesforce.com/docs/data/data-cloud-query-guide/references/dc-sql-reference/data-cloud-sql-context.html).
+This driver allows you to efficiently query large datasets in Salesforce Data Cloud with low latency and perform bulk data extractions. It's read-only, forward-only, and requires Java 8 or later. The driver uses the Data Cloud Query API SQL syntax (see [Data Cloud Query API SQL syntax](https://developer.salesforce.com/docs/data/data-cloud-query-guide/references/dc-sql-reference/data-cloud-sql-context.html) for details).
```
vogelsgesang, line 3: what does "forward-only" mean? I never heard that term before in the context of a JDBC driver?

vogelsgesang, line 3:
```suggestion
This driver allows you to efficiently query large datasets in Salesforce Data Cloud with low latency and perform bulk data extractions. It's read-only, forward-only, and requires Java 8 or later. The driver uses the [Data Cloud Query API SQL syntax](https://developer.salesforce.com/docs/data/data-cloud-query-guide/references/dc-sql-reference/data-cloud-sql-context.html).
```

I don't think we need a parenthesized sentence here, we can simply put the link directly into the sentence

README.md
```diff
@@ -1,12 +1,12 @@
...
 
+## Getting Started
 
-## Getting started
+### Maven Dependency
 
-To add the driver to your project, add the following Maven dependency:
+To use the driver in your Maven project, add the following to your `pom.xml`:
```
vogelsgesang, line 9:
```suggestion
To use the driver in your Maven project, add the following dependency to your `pom.xml`:
```

not sure, reads more natural to me that way... but I am not a native speaker, up to you, Dralia...

README.md
```diff
@@ -16,52 +16,53 @@ To add the driver to your project, add the following Maven dependency:
...
 
-Use the following command to build and test the driver:
+## Building the Driver
 
-```shell
+To build and test the driver, run:
+
+```bash
 mvn clean install
 ```
```
vogelsgesang, line 32: I think we should move this down all the way to the end of this document? Our customers should never need to build the driver from source. Building the driver is only a task which our own developers (and other open-source contributors) will have to do

README.md
```diff
@@ -16,52 +16,53 @@ To add the driver to your project, add the following Maven dependency:
...
-The class name for this driver is:
+Replace `${jdbc.version}` with the actual version of the driver. It's good practice to manage this version in a properties section of your pom.xml.
 
-```
-com.salesforce.datacloud.jdbc.DataCloudJDBCDriver
-```
+### Driver Class
 
-## Building the driver:
+The driver class name is: `com.salesforce.datacloud.jdbc.DataCloudJDBCDriver`
```
vogelsgesang, line 23: what do I need the class name for? Afaict, I don't even need to know this information for using the JDBC driver?

README.md
```diff
@@ -16,52 +16,53 @@ To add the driver to your project, add the following Maven dependency:
...
+The driver supports several OAuth authorization flows provided by Salesforce (see [OAuth authorization flows](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_flows.htm&type=5)). You'll need to configure a connected app (see [Connected App Overview](https://help.salesforce.com/s/articleView?id=sf.connected_app_overview.htm&type=5)).
 
+The following table outlines the properties for each flow:
 
-#### username and password authentication:
+| Property       | Description                    | Authentication Flow              |
+| -------------- | ------------------------------ | -------------------------------- |
+| `user`         | Salesforce username.           | Username/Password                |
+| `password`     | Salesforce password.           | Username/Password                |
+| `clientId`     | Connected app consumer key.    | All                              |
```
vogelsgesang, line 57: I think we should put the "All" entry to the top, instead of somwhere in the middle of the table

```suggestion
| `clientId`     | Connected app consumer key.    | All                              |
| `user`         | Salesforce username.           | Username/Password                |
| `password`     | Salesforce password.           | Username/Password                |
```

README.md
```diff
@@ -96,77 +95,75 @@ properties.put("clientId", "${clientId}");
...
 
-## Generated assertions
-
-Some of our classes are tested using assertions generated with [the assertj assertions generator][assertion generator].
-Due to some transient test-compile issues we experienced, we checked in generated assertions for some of our classes.
-If you make changes to any of these classes, you will need to re-run the assertion generator to have the appropriate assertions available for that class.
+Remember to replace placeholders like `${userName}` and `${query}` with your actual values. Consider using parameterized queries to prevent SQL injection vulnerabilities.
 
-To find examples of these generated assertions, look for files with the path `**/test/**/*Assert.java`.
+## Generated Assertions
```
vogelsgesang, line 158: this is only relevant for developers of the JDBC driver, not for our users.

We should probably group this together with the "Building the Driver" section

> ## Contributing to the JDBC driver
> 
> Contributions to this repository from the open-source community are welcome. For straightforward changes, feel free to just open a GIthub Pull Request. For more complex changes, we recommend to first discuss your envisioned change with us via a GIthub Issue.
> 
> ### Building the Driver
> 
> To build and test the driver, run:
> 
> ```bash
> mvn clean install
> ```
>
> ### Re-generating the Assertions
> 
> Some classes use generated assertions (from [AssertJ Assertions Generator](https://joel-costigliola.github.io/assertj/assertj-assertions-generator-maven-plugin.html#configuration)). If you modify these classes, regenerate the assertions:
> 
> ```bash
> mvn assertj:generate-assertions
> ```

vogelsgesang (approved):

README.md
```diff
@@ -16,52 +16,53 @@ To add the driver to your project, add the following Maven dependency:
 </dependency>
 ```
 
-The class name for this driver is:
+Replace `${jdbc.version}` with the actual version of the driver. It's good practice to manage this version in a properties section of your pom.xml.
 
-```
-com.salesforce.datacloud.jdbc.DataCloudJDBCDriver
-```
+### Driver Class
```
vogelsgesang, line 21: I don't think we need this section. The "Driver Class (Reiterated)" section below should be sufficient

README.md
j10t: `ResultSet` implements `getType()`, ours returns `TYPE_FORWARD_ONLY`.

Clients use this to know whether or not the ResultSet is scrollable, ours is not.

[https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#getType--](https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#getType--)

vogelsgesang: makes sense. Is "forward-only" a noteworthy enough restriction to mention it immediately in the second sentence?

To me this sounds more like an implementation detail. Most users will not really care about whether our result-sets are forward-only or not. Or am I missing something?

j10t: I’m fairly indifferent -- we can remove it, no problem. That said, I also see value in keeping it as-is since it’s a high-level functional limitation that could impact whether the driver suits an application, same as read-only. On the other hand, unlike read-only, forward-only is common among big data JDBC drivers and rarely a blocker, so most readers wouldn’t miss it if we removed it here.

Since it seemed surprisingly prominent to you here, I'd err on the side of removing it.

PR #30, opened 2025-03-20
Allow test.salesforce.com as a valid login URL for jwt token authentication

carsheng:
The intent of this fix is to allow Salesforce sandbox and test orgs to successfully use jwt authentication with the JDBC driver. 
When using the OAuth 2.0 JWT Bearer Flow, the "aud" claim should be set to the authorization server’s URL, as defined in the documentation:
Production: https://login.salesforce.com/
Sandbox: https://test.salesforce.com/
Experience Cloud: https://site.force.com/customers

Previously, loginURL test.salesforce.com resulted in a default Audience.Prod, which resulted in failed jwt authentication. Now, loginURL test.salesforce.com will default to Audience.Test which should correctly pipe through the host for authenticaiton.

### DIFF
--- a/jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/OrgIntegrationTest.java
+++ b/jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/OrgIntegrationTest.java
@@ -155,7 +155,6 @@ class OrgIntegrationTest {
     @SneakyThrows
     @ParameterizedTest
     @MethodSource("com.salesforce.datacloud.jdbc.core.StreamingResultSetTest#queryModesWithMax")
-    @Disabled
     public void exerciseQueryMode(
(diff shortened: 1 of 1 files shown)


jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/OrgIntegrationTest.java
```diff
@@ -380,6 +381,7 @@ void testMetadataUrlRoundTrip() {
 
     static boolean validateProperties() {
         AuthenticationSettings getSettings = getSettingsFromEnvironment();
+        log.info("hi");
```
jschneidereit, line 384: 😂

afurdell, line 384: 👋

jschneidereit (approved):

PR #36, opened 2025-04-03
build: Restore separate jdbc-proto artifact for better visibility

jschneidereit:
Separates proto files from jdbc-grpc classifier artifact into a dedicated jdbc-proto. This improves visibility and accessibility of the proto files for consumers who need them without requiring the full grpc implementation.

### DIFF
--- a/buildSrc/src/main/kotlin/hyper-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/hyper-conventions.gradle.kts
@@ -4,7 +4,7 @@ plugins {
 }
 
 val hyperApiVersion: String by project
-val hypzerZipPath = ".hyper/hyper-$hyperApiVersion.zip"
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/kotlin/hyper-conventions.gradle.kts
```diff
@@ -4,7 +4,7 @@ plugins {
 }
 
 val hyperApiVersion: String by project
-val hyperZipName = "hyper-$hyperApiVersion.zip"
+val hypzerZipPath = ".hyper/hyper-$hyperApiVersion.zip"
```
jschneidereit, line 7: Sneak in a tiny tweak to not put the hyper zip in `build`, which would blow it away every time you `gradle clean`

KaviarasuSakthivadivel (dismissed):

afurdell (approved):

PR #43, opened 2025-04-23
Add example that shows how to use interceptors to provide per query values.

mkaufmann:


buildSrc/build.gradle.kts
```diff
@@ -2,12 +2,6 @@ plugins {
     `kotlin-dsl`
 }
 
-java {
-    toolchain {
-        languageVersion.set(JavaLanguageVersion.of(17))
```
jschneidereit, line 7: This is here for our linting tools etc, the build still runs on Java 8, see `buildSrc/src/main/kotlin/java-conventions.gradle.kts`

PR #46, opened 2025-04-30
feature: Add Float4Vector support - Add FloatVectorAccessor, tests, and wire up in factory

KaviarasuSakthivadivel:
This change adds support for the Float4Vector data type in the JDBC driver, allowing proper handling of 32-bit floating point numbers.

### DIFF
--- a/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/FloatVectorAccessor.java
+++ b/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/FloatVectorAccessor.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.salesforce.datacloud.jdbc.core.accessor.impl;
+
+import com.salesforce.datacloud.jdbc.core.accessor.QueryJDBCAccessor;
+import com.salesforce.datacloud.jdbc.core.accessor.QueryJDBCAccessorFactory;
+import com.salesforce.datacloud.jdbc.exception.DataCloudJDBCException;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.sql.SQLException;
+import java.util.function.IntSupplier;
+import lombok.val;
+import org.apache.arrow.vector.Float4Vector;
+import org.apache.arrow.vector.holders.NullableFloat4Holder;
+
+public class FloatVectorAccessor extends QueryJDBCAccessor {
+
+    private final Float4Vector vector;
+    private final NullableFloat4Holder holder;
+
+    private static final String INVALID_VALUE_ERROR_RESPONSE = "BigDecimal doesn't support Infinite/NaN";
+
+    public FloatVectorAccessor(
+            Float4Vector vector,
+            IntSupplier currentRowSupplier,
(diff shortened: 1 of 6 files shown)


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/FloatVectorAccessor.java
```diff
@@ -0,0 +1,125 @@
...
+import lombok.val;
+import org.apache.arrow.vector.Float4Vector;
+import org.apache.arrow.vector.holders.NullableFloat4Holder;
+
+public class FloatVectorAccessor extends QueryJDBCAccessor {
+
+    private final Float4Vector vector;
+    private final NullableFloat4Holder holder;
+
+    private static final String INVALID_VALUE_ERROR_RESPONSE = "BigDecimal doesn't support Infinite/NaN";
```
jschneidereit, line 34: Should this be "float" not "BigDecimal"?

KaviarasuSakthivadivel: this error message is basically when you call getBigDecimal(), so this should be BigDecimal. 🤔

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/Float4VectorTest.java
```diff
@@ -0,0 +1,69 @@
...
+                .usePlaintext();
+
+        // Expected values from generate_series(1, 10, 0.5)
+        List<Float> expectedValues = new ArrayList<>();
+        for (float i = 1.0f; i <= 10.0f; i += 0.5f) {
+            expectedValues.add(i);
+        }
+
+        try (DataCloudConnection conn = DataCloudConnection.fromChannel(channel, properties)) {
+            try (Statement stmt = conn.createStatement()) {
```
jschneidereit, line 49: for this (managed channel and fromChannel) setup you can just use `assertWithStatement()` from HyperTestBase - we show the managed channel bit in those sample tests to be pedantic, but here we should try for brevity so the logic of the test is more clear.

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/Float4VectorTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public void testFloat4VectorQuery() throws SQLException {
+        Properties properties = new Properties();
+        ManagedChannelBuilder<?> channel = ManagedChannelBuilder.forAddress(
+                        "127.0.0.1", HyperTestBase.getInstancePort())
+                .usePlaintext();
+
+        // Expected values from generate_series(1, 10, 0.5)
+        List<Float> expectedValues = new ArrayList<>();
+        for (float i = 1.0f; i <= 10.0f; i += 0.5f) {
+            expectedValues.add(i);
```
jschneidereit, line 45: avoid mutable lists:

```java
val expectedValues = Stream.iterate(1.0f, f -> f <= 10.0f, f -> f + 0.5f)
    .collect(Collectors.toList());
```

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/Float4VectorTest.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        try (DataCloudConnection conn = DataCloudConnection.fromChannel(channel, properties)) {
+            try (Statement stmt = conn.createStatement()) {
+                ResultSet rs = stmt.executeQuery("select i::real from generate_series(1, 10, 0.5) g(i)");
+
+                int rowCount = 0;
```
jschneidereit, line 52: you could just save a list here, it makes the assert much nicer: 

```
List<Float> expectedValues = new ArrayList<>(); ugh jdbc
while (rs.next()) {
    expectedValues.add(rs.getFloat(1));
}
assertThat(allResults).containsExactlyElementsOf(expectedValues);
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/FloatVectorAccessor.java
jschneidereit: Got it 👍

jschneidereit (approved):

PR #47, opened 2025-04-30
feat: Add result set constraints and chunk-based polling

jschneidereit:
Prevent unnecessary result set iteration:
- Add `DataCloudStatement::setResultSetConstraints()` method to limit rows and bytes
- Support byte limit configuration via `resultset.bytelimit` property

Chunk-Based result polling:
- Implement waitForChunksAvailable method for pagination

Minor changes:
- Changes default byte size when using row-based access to the server's largest allowed value
- Implement `getRow()` which is optional on type forward-only
- Add examples demonstrating result set constraints with various styles of pagination
- Prefer `float4` over `real` since it's more explict


KaviarasuSakthivadivel (dismissed):

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
```diff
@@ -36,15 +37,21 @@
...
     private static final int INIT_ROW_NUMBER = -1;
 
     private final ArrowStreamReader reader;
 
-    private final AtomicInteger currentRow = new AtomicInteger(INIT_ROW_NUMBER);
+    /**
+     * Tracks the number of rows seen by this cursor, this provides the value for {@link StreamingResultSet#getRow()}
+     */
+    @lombok.Getter
+    private int row = 0;
```
afurdell, line 52: Could call it 'rowsSeen'

afurdell (dismissed):

PR #48, opened 2025-05-02
feat: add configuration for channel keepalive and retry

jschneidereit:
- Add DataCloudJdbcManagedChannel to keep original ManagedChannel clean of interceptors not provided by builder
- Refactors existing channel configuration into a consolidated area `DataCloudJdbcManagedChannel`
- Add keep-alive configuration to managed channel for connection durability
- Remove SyncQueryStatusListener and related code to reduce refactoring scope
- Removes `@Unstable` from `DataCloudStatement::getQueryId` and adds javadoc


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/HyperGrpcClientExecutor.java
```diff
@@ -281,53 +221,14 @@ private Iterator<ExecuteQueryResponse> execute(String sql, QueryParam.TransferMo
...
-
-        if (queryTimeout > 0) {
-            return result.withDeadlineAfter(queryTimeout, TimeUnit.SECONDS);
-        } else {
-            return result;
-        }
-    }
-
     private HyperServiceGrpc.HyperServiceBlockingStub getStub(@NonNull String queryId) {
         val queryIdHeaderInterceptor = new QueryIdHeaderInterceptor(queryId);
```
mkaufmann, line 231: Do we even need this method anymore on the Executor? Could we make this part of the getStub method on the new ManagedChannel?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudJdbcManagedChannel.java
```diff
@@ -0,0 +1,229 @@
...
+        configureRetries(builder, properties);
+
+        return of(builder);
+    }
+
+    private static void configureKeepAlive(ManagedChannelBuilder<?> builder, Properties properties) {
+        if (getBooleanOrDefault(properties, GRPC_KEEP_ALIVE_ENABLED, Boolean.FALSE)) {
+            val keepAliveTime = getIntegerOrDefault(properties, GRPC_KEEP_ALIVE_TIME, 60);
+            val keepAliveTimeout = getIntegerOrDefault(properties, GRPC_KEEP_ALIVE_TIMEOUT, 10);
+            val idleTimeoutSeconds = getIntegerOrDefault(properties, GRPC_IDLE_TIMEOUT_SECONDS, 1800);
```
mkaufmann, line 127: Why this very high value? Would keep it with 300s which is default in gRPC or even lower

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudJdbcManagedChannel.java
```diff
@@ -0,0 +1,229 @@
...
+    public static final String GRPC_RETRY_POLICY_MAX_BACKOFF = "grpc.retryPolicy.maxBackoff";
+    public static final String GRPC_RETRY_POLICY_BACKOFF_MULTIPLIER = "grpc.retryPolicy.backoffMultiplier";
+    public static final String GRPC_RETRY_POLICY_RETRYABLE_STATUS_CODES = "grpc.retryPolicy.retryableStatusCodes";
+
+    public HyperServiceGrpc.HyperServiceBlockingStub getStub(Properties properties, Duration queryTimeout) {
+        val interceptors =
+                configurePropertyDerivedClientInterceptors(properties).toArray(new ClientInterceptor[0]);
+
+        val timeout = queryTimeout.isZero() || queryTimeout.isNegative()
+                ? Duration.ofSeconds(Constants.DEFAULT_QUERY_TIMEOUT)
```
mkaufmann, line 73: As default I would have it unlimited, having something in JDBC that is different from server side I feel is surprising

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudJdbcManagedChannel.java
```diff
@@ -0,0 +1,229 @@
...
+    public static final String GRPC_KEEP_ALIVE_TIMEOUT = "grpc.KeepAlive.Timeout";
+    public static final String GRPC_IDLE_TIMEOUT_SECONDS = "grpc.IdleTimeoutSeconds";
+    public static final String GRPC_KEEP_ALIVE_WITHOUT_CALLS = "grpc.KeepAlive.WithoutCalls";
+
+    public static final String GRPC_RETRY_ENABLED = "grpc.enableRetries";
+    public static final String GRPC_RETRY_POLICY_MAX_ATTEMPTS = "grpc.retryPolicy.maxAttempts";
+    public static final String GRPC_RETRY_POLICY_INITIAL_BACKOFF = "grpc.retryPolicy.initialBackoff";
+    public static final String GRPC_RETRY_POLICY_MAX_BACKOFF = "grpc.retryPolicy.maxBackoff";
+    public static final String GRPC_RETRY_POLICY_BACKOFF_MULTIPLIER = "grpc.retryPolicy.backoffMultiplier";
+    public static final String GRPC_RETRY_POLICY_RETRYABLE_STATUS_CODES = "grpc.retryPolicy.retryableStatusCodes";
```
mkaufmann, line 66: Has inconsistent casing (i.e. `retryPolicy` vs `KeepAlive`)

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -225,13 +230,23 @@ public DataCloudQueryStatus waitForChunksAvailable(
...
     public DataCloudQueryStatus waitForResultsProduced(String queryId, Duration timeout) throws DataCloudJDBCException {
+        val executor = getExecutor();
         return executor.waitForResultsProduced(queryId, timeout);
     }
 
+    /**
+     * Sends a command to the server to cancel the query with the specified query id.
+     * @param queryId The query id for the query you want to cancel
+     */
+    public void cancel(String queryId) throws DataCloudJDBCException {
```
mkaufmann, line 241: Would give it more explicit name of `cancelQuery`

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -72,69 +71,71 @@ public class DataCloudConnection implements Connection, AutoCloseable {
...
+            @NonNull ManagedChannelBuilder<?> builder,
+            @NonNull Properties properties,
+            @NonNull ClientInterceptor authInterceptor,
+            @NonNull ThrowingJdbcSupplier<String> lakehouseSupplier,
+            @NonNull ThrowingJdbcSupplier<List<String>> dataspacesSupplier,
+            @NonNull DataCloudConnectionString connectionString)
+            throws DataCloudJDBCException {
+        return logTimedValue(
+                () -> DataCloudConnection.builder()
+                        .channel(DataCloudJdbcManagedChannel.of(builder.intercept(authInterceptor), properties))
```
mkaufmann, line 131: This `channel` method shoudl also explicitly have  `closeChannelWithConnection ` as parameter so that one can't set a channel but forget to set that flag. I think it's important enough to force the user to having to specify it

jdbc-util/src/main/java/com/salesforce/datacloud/jdbc/util/PropertiesExtensions.java
```diff
@@ -79,6 +83,17 @@ public static Boolean toBooleanOrDefault(String s) {
...
+        return optional(properties, key).map(PropertiesExtensions::toList).orElse(ImmutableList.copyOf(defaults));
+    }
+
+    public static List<String> toList(String s) {
+        return Arrays.stream(s.split(","))
+                .map(String::trim)
+                .filter(StringCompatibility::isNotEmpty)
+                .collect(Collectors.toList());
+    }
+
```
mkaufmann, line 96: What are these about? I'm worried that this is not handling escaping correctly (i.e. if a property value contains a comma

jschneidereit: This is to change the property `grpc.retryPolicy.retryableStatusCodes` which could be set to something like:

```
properties.setProperty("grpc.retryPolicy.retryableStatusCodes", "UNAVAILABLE, DEADLINE_EXCEEDED");
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/HyperGrpcClientExecutor.java
jschneidereit: Yeah we still need this or else higher in the stack other bits of code would need to know they need to pass a `QueryIdHeaderInterceptor` and we want to abstract that concern away from them in the "Executor" (client).

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -72,69 +71,95 @@ public class DataCloudConnection implements Connection, AutoCloseable {
...
-
-                    return DataCloudConnection.builder()
-                            .executor(executor)
-                            .properties(properties)
-                            .build();
-                },
-                "fromChannel",
+                () -> DataCloudConnection.builder()
+                        .channel(channel)
+                        .shouldCloseChannelWithConnection(closeChannelWithConnection)
```
mkaufmann, line 109: The `channel` method should force the user to directly provide the `closeChannelWithConnection` flag so that there is never any doubt and can't be missed

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -262,27 +301,29 @@ public void rollback() {}
     @Override
     public void close() {
         try {
+            if (!shouldCloseChannelWithConnection) {
+                log.warn(
+                        "Called close on a DataCloudConnection, but shouldCloseChannelWithConnection was set to false");
+                return;
+            }
```
mkaufmann, line 308: This warning doesn't make sense to me. Independent of who has channel ownership we should always ask users to close the connection as a general best practice. If there is _currently_ no other cleanup than closing the channel then we can just silently return.

There is also no reason to do this check inside the try catch, let's hoist it outside

mkaufmann (approved):

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
jschneidereit: This would happen in the case where someone did something like this:

```java
try (DataCloudConnection conn = DataCloudConnection.of(channel, properties, false)) {
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
jschneidereit: The code-generated `DataCloudConnection.Builder` is now private so it can never be instantiated by someone.
The three static `of` builder methods are the only way to build a new connection and they all have `closeChannelWithConnection`

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
jschneidereit: I think we should keep this, it basically is a warning that says "you triggered the AutoCloseable using a `try` but you didn't opt-in to the functionality of the automatic cleanup with `true`".

I did lift the warn out of the try/catch though

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
jschneidereit: Same as below, there's no way to forget to set the flag now

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
mkaufmann: >I think we should keep this, it basically is a warning that says "you triggered the AutoCloseable using a try but you didn't opt-in to the functionality of the automatic cleanup with true".

Let's make it a debug / trace log. I'm not a fan of issuing a warning when the client is doing everything correctly. The intent looks to be more like a trace / debug hint

mkaufmann (approved):

PR #54, opened 2025-05-10
feat: Implement basic Spark Data Source

vogelsgesang:
This commit adds a basic Spark data source for reading query results from Hyper. It can be used as

```
spark.read
  .format("com.salesforce.datacloud.spark.HyperResultSource")
  .option("query_id", queryId)
  .option("port", 8585)
  .load()
```

Functionality
-------------

The Spark source uses Hyper's chunked results for distributed processing, fetching to fetch different chunks on different Spark workers.

We currently do provide the row count statistics to Spark optimizer, but no advanced statistics, yet.

The driver provides custom metrics (row count and chunk count) which can be observed in Spark's metric framework.

All the new code lives in the new `spark-datasource` subproject.

Future work:
* The connection setup is still very basic, and only supports unencrypted, IP-based connection setup. This will have to evolve, such that the full set of JDBC options are available also via the Spark driver. This will likely require cleaning up the option handling of the JDBC driver itself first, though.
* NULLs are not supported, yet. Unfortunately, the JDBC currently reports nullability of columns incorrectly.
* Not all data types are read correctly, yet. E.g., the types of Numeric columns are reported incorrectly. This is due to a bug in the underlying JDBC library. Also, array support is still missing.

Build system changes
--------------------

The `spark-datasource` subproject is written in Scala, as is the rest of Spark. Hence, the build system needed a couple of small changes, to also accomodate Scala code. Among others, I did add `scalafmt` to spotless.

I wanted to reuse the `HyperServerProcess` utility also in my new test cases. To make those classes available across subprojects, I had to move them into the `testFixtures` package.

Future work:
* Split into `spark-datasource` and `spark-datasource-core`. Currently, the `spark-datasource` driver also contains grpc. This might cause issues if other grpc library versions are already loaded in the Spark Java VM host. We need to split up the packages to solve this issue, similar to how we did it for `jdbc-core` vs `jdbc`.


spark-datasource/src/main/scala/com/salesforce/datacloud/spark/HyperResultTable.scala
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  override def newScanBuilder(
+      options: CaseInsensitiveStringMap
+  ): ScanBuilder = {
+    val (chunkCount, rowCount) =
+      Using(connectionOptions.createConnection()) { conn =>
+        // TODO XXX: set infinite timeout
+        val queryStatus =
+          conn.waitForResultsProduced(resultSetId, Duration.ofDays(2))
```
jakeswenson, line 38: Should this timeout be configurable? I can't imagine why infinite/2days would be a good thing for a spark query

spark-datasource/src/test/scala/HyperResultSourceTest.scala
```diff
@@ -0,0 +1,175 @@
...
+    val rows = spark.read
+      .format("com.salesforce.datacloud.spark.HyperResultSource")
+      .option("port", hyperServerProcess.getPort())
+      .option("query_id", queryId)
+      .load()
+      .collect()
+
+    // Verify that we read all rows.
+    assert(rows.length == 1000)
+
```
jakeswenson, line 159: doesn't hurt to add an assertion that the result is a sequence (sum the results and compare with expected?) i.e. just test reading multiple rows correctly.

jakeswenson (approved):

jdbc-core/build.gradle.kts
```diff
@@ -1,5 +1,6 @@
 plugins {
     id("java-conventions")
+    id("java-test-fixtures")
```
jschneidereit, line 3: This should be made unnecessary by #72

spark-datasource/build.gradle.kts
```diff
@@ -0,0 +1,32 @@
+plugins {
+    id("scala")
+    id("java-conventions")
```
jschneidereit, line 3: Should `java-conventions` be here? We should figure out what is needed from that and move it to `scala-conventions`

vogelsgesang: would it rather make sense to move this into a `code-formatting.gradle.kts` file which would cover all code formatting (Java, Scala and "gitattributes")?

Having gitattributes formatting in the Java conventions is weird, anyway...

Is there a way to apply a *.gradle.kts globally, across all sub-projects? Preferably, the code-formatting.gradle.kts would be completely global.

jdbc-core/build.gradle.kts
vogelsgesang: makes sense, let's see which one lands first

jschneidereit: Yeah `HyperResultSourceTest.scala` would need to implement a small shim, this is because #72 tries to avoid adding jdbc-core concept to what should just be the hyperd management code

praveen2450 (approved):

PR #72, opened 2025-06-18
Move HyperServerProcess and friends to hyper-test-util submodule

jschneidereit:
This PR refactors `HyperServerProcess` into its own submodule. This will make it possible for other submodules to easily spin up new instances of hyper without having to deal with weirdness from test fixtures and makes more sense in a monorepo like this.

To use just add `testImplementation(project(":hyper-test-util"))` to your dependencies block of your build.gradle.kts file.


jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/hyper/HyperConnectionAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Map;
+import java.util.Properties;
+import lombok.AllArgsConstructor;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+import lombok.val;
+
+@AllArgsConstructor
+@Slf4j
+public class HyperConnectionAdapter implements AutoCloseable {
```
vogelsgesang, line 37: why do we need this class now? Which purpose does it serve? The name `HyperConnectionAdapter` sounds as if it would adopt one type of connections to some other type of connection, but this is apparently not what it is doing. The absence of a comment describing when / how to use this class doesn't help, either

hyper-test-util/src/main/java/com/salesforce/datacloud/hyper/HyperServerProcess.java
```diff
@@ -133,25 +137,4 @@ public void close() throws Exception {
...
-    }
-
-    @SneakyThrows
-    public DataCloudConnection getConnection(Map<String, String> connectionSettings) {
-        val properties = new Properties();
-        properties.put(DirectDataCloudConnection.DIRECT, "true");
-        properties.putAll(connectionSettings);
-        val url = CONNECTION_PROTOCOL + "//127.0.0.1:" + getPort();
-        return DirectDataCloudConnection.of(url, properties);
-    }
```
vogelsgesang, line 156: removing those function will make reuse much harder.

E.g., I couldn't would still need to use testFixtures to expose the HyperConnectionAdapter class, so I could use this from my Scala tests in #54

hyper-test-util/src/main/java/com/salesforce/datacloud/hyper/HyperServerProcess.java
```diff
@@ -133,25 +137,4 @@ public void close() throws Exception {
         log.warn("shutdown hyper monitors");
         hyperMonitors.shutdown();
     }
```
mkaufmann, line 139: We should also explicitly clean up the temp file here, so that they don't just get deleted at process end

hyper-test-util/src/main/java/com/salesforce/datacloud/hyper/HyperServerProcess.java
```diff
@@ -101,11 +90,26 @@ public HyperServerProcess(HyperServerConfig.HyperServerConfigBuilder config) {
...
 
+    private Path extractHyperYamlToTempFile() throws IOException {
+        val tempFile = Files.createTempFile("hyper", ".yaml");
+        tempFile.toFile().deleteOnExit();
+
+        try (val inputStream = HyperServerProcess.class.getResourceAsStream("/hyper.yaml")) {
+            if (inputStream == null) {
+                throw new IllegalStateException("Could not find hyper.yaml resource in classpath");
+            }
+            Files.copy(inputStream, tempFile, StandardCopyOption.REPLACE_EXISTING);
```
mkaufmann, line 101: REPLACE_EXISTING sounds wrong, another process might have a different config and this can lead to surprising behavior. It should rather fail on conflicts

hyper-test-util/src/main/java/com/salesforce/datacloud/hyper/HyperServerProcess.java
mkaufmann: This module provides the basic process functionality without having a dependency on `jdbc-core`. Thus it can't provide utility methods for `DataCloudConnection`. That is still covered by `HyperConnectionAdapter` (which I think needs a better name). Overall the PR preserves this functionality in that class.

I'm not sure though what the motivation was to do this PR at this point, if it was only for your scala change I guess having the code in a separate module feels premature

hyper-test-util/build.gradle.kts
```diff
@@ -0,0 +1,11 @@
+plugins {
+    id("java-conventions")
+    alias(libs.plugins.lombok)
+}
+
+dependencies {
+    implementation(platform(libs.junit.bom))
+    implementation(libs.bundles.testing)
+    implementation(libs.jackson.databind)
+    implementation(libs.guava)
```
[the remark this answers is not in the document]
jschneidereit: junit.bom controls the versions of junit pulled in by bundles.testing, and one of the classes implements `BeforeAllCallback`. As for Guava we use that for immutable lists and such because of java 8 🥲

hyper-test-util/src/main/java/com/salesforce/datacloud/hyper/HyperServerProcess.java
vogelsgesang: I think the "test fixtures" from the Spark review is the actual better approach then.

I don't see which value this PR provides over just using test fixtures.

vogelsgesang: > I'm not sure though what the motivation was to do this PR at this point, if it was only for your scala change I guess having the code in a separate module feels premature

This.

@jschneidereit please provide the motivation why you think we should do this change here

PR #73, opened 2025-06-23
fix: apply refactor > delombok to remove @UtilityClass

jschneidereit:
I suspect that there's an issue with the way `@UtilityClass` works and can cause problems with Spark at the very least. It took about 60 seconds to use https://plugins.jetbrains.com/plugin/6317-lombok to remove all instances of it.


vogelsgesang (commented):
> I suspect that there's an issue with the way @UtilityClass works and can cause problems with Spark at the very least.

Please provide more details on what the observed issue was, and why we think that this change might fix it

mkaufmann (approved):

carsheng (approved):

PR #77, opened 2025-07-03
debt: Refactor Property Handling

mkaufmann:
In preparation for cleaning up the various timeout handlings in the JDBC driver this change first cleans up the property handling.
All properties will now be interpreted when the connection is created instead of the previous approach where it was more distributed at the places where the respective properties are used.
For queryTimeout this for example had the benefit that it's much easier to understand how it is computed and the internal code doesn't need to check for invalid user inputs again and again.

The settings are structured into sub classes that coarsely represent the area for which they are relevant.
This structuring helped to for example identify that `resultset.bytelimit` was used at the wrong levels (see notes further down).
The change also introduces the `PropertiesRegistry` file that easily allows to add handling for new properties.
For now the registry only covers properties relevant for the Connection & Statement objects. Future refactorings should also make use of it for the Authorization related properties.

With the transition of properties to this new framework we also removed the "silent" ignoring of malformed property values.
Just ignoring property values for valid keys can lead to surprising behavior like for example timeouts getting ignored, thus explicitly failing is beneficial.

Removed Properties:
- The `resultset.bytelimit` property was removed. It didn't work for usage in DBeaver or other generic JDBC clients anyway as it would only have effect when the max rows are limited. As currently it is sufficient that these settings can be configured through our API extensions instead of fixing and adding more properties for maxRows this change removes the property.

Related Changes:
- Double enforcement of query timeout on stub and in the executor: The query timeout was used both for individual generic RPC requests at the connection level as well as governing the whole end to end query execution flow encompassing multiple RPC requests. To clean this up the individual RPC call timeout is now detached from the queryTimeout and instead associated with the networkTimeout (with the standardized get / set support).
- Don't keep `byteLimit` as a member in `HyperGprcClientExecutor`. The `byteLimit` is only relevant when there is a row limit. In the existing code the byteLimit was set as a member during construction for all use cases while the row limit was only passed as argument in specific `executeQuery` overloads. This split was highly confusing. The `byteLimit` is now explicitly passed in on the methods that need it alongside the row limit.
- This change fixes the behavior of the `client info` related connection methods. Our driver currently doesn't support client info properties and thus we just return an empty properties set / null when returning those and ignore set values. This behavior mirrors the documented behavior of the MSSQL driver.
- Removed support for legacy `serverSetting.` prefix

The actual proper timeout handling alignment and cleanup will be done separately in a follow up.


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -48,21 +43,26 @@ public class DataCloudStatement implements Statement, AutoCloseable {
...
+    private StatementProperties statementProperties;
+
+    // The target maximum number of rows for a query. The default means disabled.
+    @Getter(AccessLevel.PACKAGE)
+    private int targetMaxRows = 0;
+
+    // The target maximum number of bytes per RPC response. This is only relevant when `targetMaxRows > 0`.
+    // The default means disabled.
+    @Getter(AccessLevel.PACKAGE)
+    private int targetMaxBytes = 0;
```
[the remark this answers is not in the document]
mkaufmann: Thanks, adjusted that here and also in the `...Properties` classes

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -209,33 +211,42 @@ public void setMaxRows(int max) {}
...
+        return (int) statementProperties.getQueryTimeout().getSeconds();
     }
 
+    /**
+     * Sets the query timeout for this statement. A zero or negative value is interpreted as infinite timeout.
+     * @param seconds The query timeout in seconds.
+     */
     @Override
     public void setQueryTimeout(int seconds) {
+        // Zero or negative values are interpreted as infinite timeout
```
jschneidereit, line 243: Should we remove this first line in the comment since it's redundant with the javadoc?

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/PropertiesTest.java
```diff
@@ -31,35 +33,67 @@
...
+        // Cover string setting
+        properties.setProperty("workload", "testWorkload");
+        // Cover prefixed setting
+        properties.setProperty("querySetting.A", "A");
+        properties.setProperty("querySetting.B", "B");
+        // Cover setting that internally is represented as non string (`Duration`)
+        properties.setProperty("queryTimeout", "30");
+        ConnectionProperties connectionProperties = ConnectionProperties.of(properties);
+        Properties roundtripProperties = connectionProperties.toProperties();
+        assertThat(roundtripProperties.entrySet()).containsExactlyInAnyOrderElementsOf(roundtripProperties.entrySet());
```
jschneidereit, line 87: This assertion looks off, I'd expect the `expected` to be from a different object than the `actual`

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StatementProperties.java
```diff
@@ -0,0 +1,105 @@
...
+                Duration timeout = Duration.ofSeconds(Integer.parseInt(queryTimeoutStr));
+                if (timeout.isZero() || timeout.isNegative()) {
+                    builder.queryTimeout(Duration.ZERO);
+                } else {
+                    builder.queryTimeout(timeout);
+                }
+            } catch (NumberFormatException e) {
+                throw new DataCloudJDBCException("Failed to parse queryTimeout: " + e.getMessage());
+            }
+        }
```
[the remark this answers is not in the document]
mkaufmann: My feeling was to have it all in a single method so that one cane have a single look to understand all parsing logic without having to jump around in the file.
I feel that for methods that are <100-200 lines of code I try to optimize for locality (especially as there is no duplication here) and only once it gets bigger I check if there are good sized blocks that can be outlined.

If you feel that the complexity of the method overall is still acceptable I would keep as is.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -539,12 +549,25 @@ public String getSchema() {
...
+     * To enforce a query timeout rather use {@link DataCloudStatement#setQueryTimeout(int)}.
+     * A too low network timeout might cause the JDBC driver to fail to operate properly.
+     * @param executor This will be ignored
+     * @param milliseconds The network timeout in milliseconds.
+     */
     @Override
-    public void setNetworkTimeout(Executor executor, int milliseconds) {}
+    public void setNetworkTimeout(Executor executor, int milliseconds) {
+        networkTimeout = Duration.ofMillis(milliseconds);
+    }
```
vogelsgesang, line 562: I would prefer if the user can pass this in directly, as a strictly typed `Duration`.

Also, why do we not accept this as part of `ConnectionProperties`? Might be useful to pass this in via the JDBC property bag, too, e.g. from DBeaver for quick experiments?

```suggestion
    public void setNetworkTimeout(Executor executor, Duration timeout) {
        networkTimeout = timeout;
    }
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -539,12 +549,25 @@ public String getSchema() {
...
 
+    /**
+     * Returns the network timeout for this connection.
+     * @return The network timeout for this connection in milliseconds.
+     */
     @Override
     public int getNetworkTimeout() {
```
vogelsgesang, line 569:
```suggestion
    public Duration getNetworkTimeout() {
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -152,49 +151,41 @@ public static DataCloudConnection of(
...
-                    "Built stub with queryTimeout={}, interceptors={}",
-                    queryTimeout,
-                    metadata.keys().size());
-            stub = stub.withDeadlineAfter(queryTimeout.getSeconds(), TimeUnit.SECONDS);
-        } else {
-            log.info(
-                    "Built stub with queryTimeout=none, interceptors={}",
-                    metadata.keys().size());
+        if (!networkTimeout.isZero()) {
+            stub = stub.withDeadlineAfter(networkTimeout.toMillis(), TimeUnit.MILLISECONDS);
```
vogelsgesang, line 164: I am a bit confused about the semantics of this deadline here.

From internal discussions, I understood you wanted this to have the semantics

> networkTimeout - Restricts the runtime of individual RPC requests, can be used as a failsafe to protect against random network hangs. 

However, this is not what `withDeadlineAfter` does. It sets a general timeout for *all* RPC requests together. And the time starts counting right now, after you called `withDeadlineAfter`, even if the stub would be unused for a couple more seconds / minutes before sending the first RPC request

I.e., the timer starts running with the creation of the `DataCloudStatement`, even if the query isn't even submitted, yet. That's rather unexpected IMO

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/partial/RowBased.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.salesforce.datacloud.jdbc.core.partial;
 
+import com.salesforce.datacloud.jdbc.core.DataCloudStatement;
```
vogelsgesang, line 18: layering-wise, this dependency feels a bit backwards. Not sure waht `RowBased` does exactly, but it feels like this should live a layer below `DataCloudStatement`?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -48,21 +43,30 @@ public class DataCloudStatement implements Statement, AutoCloseable {
...
+     */
+    @Getter(AccessLevel.PACKAGE)
+    private int targetMaxBytes = 0;
 
     public DataCloudStatement(@NonNull DataCloudConnection connection) {
         this.connection = connection;
-        val properties = connection.getClientInfo();
-        this.queryTimeout = getIntegerOrDefault(properties, QUERY_TIMEOUT, DEFAULT_QUERY_TIMEOUT);
-        this.targetMaxBytes = getIntegerOrDefault(properties, BYTE_LIMIT, HYPER_MAX_ROW_LIMIT_BYTE_SIZE);
+        statementProperties = connection.getConnectionProperties().getStatementProperties();
```
vogelsgesang, line 64: for consistency with line above:

```suggestion
        this.statementProperties = connection.getConnectionProperties().getStatementProperties();
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -15,16 +15,11 @@
  */
 package com.salesforce.datacloud.jdbc.core;
 
-import static com.salesforce.datacloud.jdbc.core.HyperGrpcClientExecutor.HYPER_MAX_ROW_LIMIT_BYTE_SIZE;
-import static com.salesforce.datacloud.jdbc.core.HyperGrpcClientExecutor.HYPER_MIN_ROW_LIMIT_BYTE_SIZE;
```
vogelsgesang, line 19: why were those moved here? Layering-wise, they felt more natural in the `HyperGrpcClientExecutor` to me

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -150,17 +153,30 @@ public int getMaxFieldSize() {
...
     public void clearResultSetConstraints() throws DataCloudJDBCException {
         setResultSetConstraints(0, HYPER_MAX_ROW_LIMIT_BYTE_SIZE);
     }
 
-    private void checkResultSetConstraints(int maxRows, int maxBytes) throws DataCloudJDBCException {
+    /**
+     * Sets the constraints that would limit the number of rows and overall bytes in any ResultSet object generated by this Statement.
+     * This is used to tell the database the maximum number of rows and bytes to return, you may get less than expected because of this.
+     *
+     * @param maxRows The target maximum number of rows a ResultSet can have, zero means there is no limit.
```
vogelsgesang, line 173: Afaict, we will never return more than those?

```suggestion
     * @param maxRows The maximum number of rows a ResultSet can have, zero means there is no limit.
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -209,33 +211,42 @@ public void setMaxRows(int max) {}
...
-        return queryTimeout;
+        return (int) statementProperties.getQueryTimeout().getSeconds();
     }
 
+    /**
+     * Sets the query timeout for this statement. A zero or negative value is interpreted as infinite timeout.
+     * @param seconds The query timeout in seconds.
+     */
     @Override
     public void setQueryTimeout(int seconds) {
```
vogelsgesang, line 242: Rather use proper types

```suggestion
    public void setQueryTimeout(Duration seconds) {
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -209,33 +211,42 @@ public void setMaxRows(int max) {}
...
-    protected Duration getQueryTimeoutDuration() {
-        return Duration.ofSeconds(getQueryTimeout());
+        return statementProperties.getQueryTimeout();
     }
 
+    /**
+     * @return The query timeout for this statement in seconds.
+     */
     @Override
     public int getQueryTimeout() {
```
vogelsgesang, line 233:
```suggestion
    public Duration getQueryTimeout() {
```

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/PropertiesTest.java
```diff
@@ -31,35 +33,67 @@
...
+                statement -> {
+                    val result = statement.executeQuery("SHOW date_style");
+                    result.next();
+                    assertThat(result.getString(1)).isEqualTo("ISO, YMD");
+                },
+                settings);
+    }
+
+    @Test
+    void testGetSettingWithCorrectPrefix() throws DataCloudJDBCException {
```
vogelsgesang, line 54: here and other test cases in this file: I am having a hard time to understand the intended test coverage. Can you please rename the test cases / add comments to more clearly articulate the intended test coverage?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StatementProperties.java
```diff
@@ -0,0 +1,105 @@
...
+     */
+    public static StatementProperties of(Properties props) throws DataCloudJDBCException {
+        StatementPropertiesBuilder builder = StatementProperties.builder();
+
+        // The query timeout property, zero or negative values are interpreted as infinite timeout.
+        // Positive values are interpreted as the number of seconds for the timeout
+        String queryTimeoutStr = props.getProperty("queryTimeout");
+        if (queryTimeoutStr != null) {
+            try {
+                Duration timeout = Duration.ofSeconds(Integer.parseInt(queryTimeoutStr));
```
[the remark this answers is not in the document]
mkaufmann: That was the existing behavior, there is a separate thread / work for adjusting the queryTimeout setting to support decimals and potentially interval syntax. I see that out of scope for this PR

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
mkaufmann: This signature is mandated by the standard

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ConnectionProperties.java
```diff
@@ -0,0 +1,115 @@
...
+     * The workload to use for the connection (default: jdbcv3)
+     */
+    @Builder.Default
+    private final String workload = "jdbcv3";
+
+    /**
+     * The external client context to use for the connection
+     */
+    @Builder.Default
+    private final String externalClientContext = "";
```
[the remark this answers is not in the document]
mkaufmann: I checked (and now also added test coverage) that the user gets an actionable error message when submitting an invalid JSON.
Given that I think not checking in the driver is the right approach to not accidentally create diverging behavior in what is accepted or not.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
mkaufmann: Great catch, this was an area where I skimped on test coverage. I've started with adding test coverage first now. For  I added a small utility class to interact with the hyper log to verify using the server side logs that the right thing is happening.

There was indeed a problem before and the new approach of injecting a per call deadline fixes the issue

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
vogelsgesang: Given that this is not a regression, feel free to ship this PR as is - or fix in the same PR. Just let me know what you prefer to do :)

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/PropertiesTest.java
vogelsgesang: Thanks! Looks good now

carsheng (approved):

PR #80, opened 2025-07-09
feat: Improved query timeout handling

mkaufmann:
This change adjusts the query timeout handling from being locally  + gRPC deadline enforced to now primarily being enforced on Hyper server side.
- This allows to return the server side error to the caller. Previously it would either raise an internal error by the JDBC driver or obscure gRPC timeout errors depending on which layer detected the problem first.
- It now also guarantees that async queries don't exceed their timeout even if the client crashes / or the user flow forgets to explicitly invoke `cancel`.

To ensure robustness of the query timeout handling in face of network problems or even server side bugs this change doesn't only rely on the server side timeouts.
As a safety net there is also local enforcement (which happens with a configurable delay to allow the server to terminate first)

As a related change the driver now also internally tracks timeouts more accurately (both for executeQuery as well as for the custom `wait*` methods).
- The old approach was in some places relying on non monotonic clocks (`Instant`) and thus would have resulted in failures around system clock updates / summer time changes
- The old approach sometimes waited for a multiple of the configured query timeout

To remove this class of issues usages a new Deadline class is introduced that tracks the remaining time for a timeout.

### DIFF
--- a/jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/QueryTimeoutTest.java
+++ b/jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/QueryTimeoutTest.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2024, Salesforce, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.salesforce.datacloud.jdbc.core;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
+
+import com.salesforce.datacloud.jdbc.exception.DataCloudJDBCException;
+import com.salesforce.datacloud.jdbc.hyper.HyperTestBase;
+import com.salesforce.datacloud.jdbc.util.HyperLogScope;
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.time.Duration;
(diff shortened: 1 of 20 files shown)


jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/listener/QueryStatusListenerTest.java
```diff
@@ -38,9 +39,9 @@ private QueryStatusListener sut(String query, QueryParam.TransferMode mode) {
     private QueryStatusListener sut(String query, QueryParam.TransferMode mode, Duration timeout) {
         switch (mode) {
             case ASYNC:
-                return AsyncQueryStatusListener.of(query, hyperGrpcClient, timeout);
+                return AsyncQueryStatusListener.of(query, hyperGrpcClient, QueryTimeout.of(timeout, Duration.ZERO));
             case ADAPTIVE:
-                return AdaptiveQueryStatusListener.of(query, hyperGrpcClient, timeout);
+                return AdaptiveQueryStatusListener.of(query, hyperGrpcClient, QueryTimeout.of(timeout, Duration.ZERO));
```
jschneidereit, line 44: Should we make an overload of `QueryTimeout.of` where the local enforcement delay defaults to `Duration.ZERO`?

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/QueryTimeoutTest.java
```diff
@@ -0,0 +1,147 @@
...
+                statement.executeQuery("SHOW query_timeout");
+            }
+        }
+
+        // Verify that the gRPC deadline is far below the 100 seconds that the network timeout is set to.
+        ResultSet resultSet = logScope.executeQuery(
+                "select CAST(v->>'requested-timeout' as DOUBLE PRECISION)from hyper_log WHERE k='grpc-query-received'");
+        resultSet.next();
+        assertThat(resultSet.getDouble(1)).isLessThan(20);
+        logScope.close();
```
jschneidereit, line 118: Does it make sense to use `try` semantics for this in case the test fails?

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/listener/QueryStatusListenerTest.java
mkaufmann: Such an overload would only be required for test. For normal production code it would be dangerous as the timeout should always be passed from another source. To keep the complexity for writing the driver code low I wouldn't add such overload and pay the small fee in tests.

PR #81, opened 2025-07-11
feat: Allow toggling of including query details in exception messages

mkaufmann:
For some use cases like logging in an app server it can be undesirable to log query / snippets as they might contain data that should not be logged. This change adds new property `errorsIncludeCustomerDetails` that allows removing such information from the generic exception message. The default is kept like before to still show it.
Independent of the mode the full information is preserved in the exception and can be access using `getFullCustomerMessage()`.

As the change needs to no support having the query included / excluded from the message the format was adjusted to ensure such an inclusion/exclusion doesn't impact the readability of the message (same also for query id, which might be available in some context and missing in others). The new general error formatting is
```
Failed to execute query: <server side message / local exception message>
SQLSTATE: <server provided / "HY==" for generic error>
QUERY-ID: ... <-- Line only if available
DETAIL: ... <-- Line only if available
HINT:... <-- Line only if available
QUERY: ... <-- Line only if available
```

With this change the signature for exception creation was adjusted to explicitly required always passing `query-id` and `query`. There were many places where the query id could have been passed but wasn't and same for the query. (Not all places were fixed in this PR because that would require larger refactorings).

Drive-By:
- Remove unused signatures and their test coverage
- Fix missing serialization of `queryTimeoutLocalEnforcementDelay` property

Future work:
- The new interface uncovered several places on the execute query flow where the query is not available for formatting the exception. Those places should be adjusted so that they get access to the query string
- The error info contains additional information that is not yet part of the exception. We should revisit our exception to add this information as well.
- Our `DataCloudJDBCException` has many overloads that just forward to the generic SQLException. Those could get removed and those places could just throw an SQLException

### DIFF
--- a/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
+++ b/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
@@ -67,7 +67,10 @@ public class StreamingResultSet extends AvaticaResultSet implements DataCloudRes
 
     @SneakyThrows
     public static StreamingResultSet of(
-            String queryId, HyperGrpcClientExecutor client, Iterator<QueryResult> iterator) {
+            String queryId,
+            HyperGrpcClientExecutor client,
+            Iterator<QueryResult> iterator,
+            boolean includeCustomerDetailInReason) {
         try {
             val channel = ExecuteQueryResponseChannel.of(StreamUtilities.toStream(iterator));
             val reader = new ArrowStreamReader(channel, new RootAllocator(ROOT_ALLOCATOR_MB_FROM_V2));
@@ -85,7 +88,7 @@ public class StreamingResultSet extends AvaticaResultSet implements DataCloudRes
 
             return result;
         } catch (Exception ex) {
-            throw createException(QUERY_FAILURE + queryId, ex);
+            throw createException(null, queryId, includeCustomerDetailInReason, ex);
         }
     }
 
@@ -102,8 +105,6 @@ public class StreamingResultSet extends AvaticaResultSet implements DataCloudRes
 
         return client.getQueryStatus(queryId);
     }
-
-    private static final String QUERY_FAILURE = "Failed to execute query: ";
 
     @Override
     public int getType() {
--- a/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandler.java
+++ b/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandler.java
@@ -19,8 +19,9 @@ import com.google.protobuf.Any;
 import com.google.protobuf.InvalidProtocolBufferException;
 import io.grpc.StatusRuntimeException;
 import io.grpc.protobuf.StatusProto;
-import java.sql.SQLException;
 import java.util.List;
+import javax.annotation.Nullable;
+import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import salesforce.cdp.hyperdb.v1.ErrorInfo;
 
@@ -34,14 +35,34 @@ public final class QueryExceptionHandler {
         throw new UnsupportedOperationException("This is a utility class and cannot be instantiated");
     }
 
-    public static DataCloudJDBCException createQueryException(String query, Exception e) {
-        String exceptionQuery = query.length() > MAX_QUERY_LENGTH_IN_EXCEPTION
-                ? query.substring(0, MAX_QUERY_LENGTH_IN_EXCEPTION) + "<truncated>"
-                : query;
-        return QueryExceptionHandler.createException("Failed to execute query: " + exceptionQuery, e);
-    }
+    /**
+     * Creates an exception for a specific query id.
+     *
+     * @param query The query that failed (can be null for scenarios like fetching by query id)
+     * @param queryId The query id that failed (can be null in case execution failed before request reaches the server)
+     * @param includeCustomerDetailInReason Whether to include customer details in the exception message (i.e. fragments of the query that can contain strings provided by the customer)
+     * @param e The exception that occurred.
+     * @return A DataCloudJDBCException with the query id and the exception message.
+     */
+    public static DataCloudJDBCException createException(
+            @Nullable String query,
+            @Nullable String queryId,
+            boolean includeCustomerDetailInReason,
+            @NonNull Exception e) {
+        // Ensure query is truncated if it's too long.
+        if (query != null) {
+            query = query.length() > MAX_QUERY_LENGTH_IN_EXCEPTION
+                    ? query.substring(0, MAX_QUERY_LENGTH_IN_EXCEPTION) + "<truncated>"
+                    : query;
+        }
 
-    public static DataCloudJDBCException createException(String message, Exception e) {
+        // This is a generic error SQL state, we'll use this if we don't have a specific SQL state from the error info
+        String sqlState = "HY000";
+        String grpcMessage = null;
+        String primaryMessage = null;
+        String customerHint = null;
+        String customerDetail = null;
+
         if (e instanceof StatusRuntimeException) {
             StatusRuntimeException ex = (StatusRuntimeException) e;
             com.google.rpc.Status status = StatusProto.fromThrowable(ex);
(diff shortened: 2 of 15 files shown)


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -67,7 +67,10 @@ private StreamingResultSet(
 
     @SneakyThrows
     public static StreamingResultSet of(
-            String queryId, HyperGrpcClientExecutor client, Iterator<QueryResult> iterator) {
+            String queryId,
+            HyperGrpcClientExecutor client,
+            Iterator<QueryResult> iterator,
+            boolean includeCustomerDetailInReason) {
```
vogelsgesang, line 73: do we foresee the need for the `StreamingResultSet` to access other properties from the `StatementProperties`, too? Should we forward the whole `StatementProperties`?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandler.java
```diff
@@ -57,31 +78,65 @@ public static DataCloudJDBCException createException(String message, Exception e
...
+        StringBuilder sharedMessageBuilder = new StringBuilder("Failed to execute query");
+        // We use status.getMessage() as in contrast to the primary message this already has the trace id
+        // encoded.
+        if (grpcMessage != null) {
+            sharedMessageBuilder.append(": " + grpcMessage);
+        } else {
+            // If we don't have any information from the server we resort to using the exception message
+            sharedMessageBuilder.append(": " + e.getMessage());
+        }
+        sharedMessageBuilder.append(String.format("%nSQLSTATE: %s", sqlState));
```
vogelsgesang, line 102: use modern string interpolation instead of `String.format`

Here and other places

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandler.java
```diff
@@ -34,14 +35,34 @@ private QueryExceptionHandler() {
...
+    /**
+     * Creates an exception for a specific query id.
+     *
+     * @param query The query that failed (can be null for scenarios like fetching by query id)
+     * @param queryId The query id that failed (can be null in case execution failed before request reaches the server)
+     * @param includeCustomerDetailInReason Whether to include customer details in the exception message (i.e. fragments of the query that can contain strings provided by the customer)
+     * @param e The exception that occurred.
+     * @return A DataCloudJDBCException with the query id and the exception message.
+     */
+    public static DataCloudJDBCException createException(
```
vogelsgesang, line 47: why is this in a separate utility class? Shouldn't this be package-private inside the `DataCloudJDBCException` class?

jdbc-util/src/main/java/com/salesforce/datacloud/jdbc/exception/DataCloudJDBCException.java
```diff
@@ -20,14 +20,14 @@
 
 @Getter
 public class DataCloudJDBCException extends SQLException {
+    private String fullCustomerMessage;
+
+    private String primaryMessage;
+
     private String customerHint;
 
     private String customerDetail;
```
vogelsgesang, line 29: please add comment on what those represent and when they should be used. Get inspiration from the corresponding protobuf files

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandlerTest.java
vogelsgesang, line 1: use text blocks (https://docs.oracle.com/en/java/javase/15/text-blocks/index.html) for all multi-line strings

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandlerTest.java
```diff
@@ -15,65 +15,140 @@
...
+                        // characters for the double quotes)
+                        + "line 1, column 38: WITH \"A\" AS (SELECT 1) SELECT * FROM A\n"
+                        + "                                                        ^";
+                String customerHint = "Try quoting the identifier: `\"A\"`";
+                String expectedMessage = "Failed to execute query: table \"a\" does not exist\n" + "SQLSTATE: 42P01\n"
+                        + "QUERY-ID: "
+                        + stmt.getQueryId() + "\n" + "DETAIL: "
+                        + customerDetail
+                        + "\n" + "HINT: "
+                        + customerHint;
```
vogelsgesang, line 54: the placement of the `"\n"` newlines is pretty arbitrary...
Can we use string interpolation/string templates + text blocks?

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandlerTest.java
```diff
@@ -15,65 +15,140 @@
...
-    public void testCreateExceptionWithSQLStateAndMessage() {
-        String mockSQLState = "42P01";
-        SQLException sqlException = QueryExceptionHandler.createException("test message", mockSQLState);
+    void testCreateExceptionWithGenericException() {
+        Exception mockException = new Exception("Host not found");
+        String fullMessage = "Failed to execute query: Host not found\n" + "SQLSTATE: HY000\n"
+                + "QUERY-ID: 1-2-3-4\n"
+                + "QUERY: SELECT 1";
+        String redactedMessage =
+                "Failed to execute query: Host not found\n" + "SQLSTATE: HY000\n" + "QUERY-ID: 1-2-3-4";
```
vogelsgesang, line 135: weird formatting...

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandlerTest.java
mkaufmann: I wanted to do that originally but it doesn't work for our Java 8 build

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandler.java
mkaufmann: The DataCloudJDBCException is in the jdbc-util package which doesn't have a dependency on any gRPC related method. Thus to move the method there we would need to move the DataCloudJDBCException to the jdbc-core module. I saw that as going too far for this PR and see that more as in scope for the general JDBC quality uplift.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
mkaufmann: I don't know yet, for now I prefer to explicitly pass the required information to help at the call site to more easily see what is getting used. I had trouble in multiple places in our driver to understand which properties are used because the full properties object was passed on.

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/exception/QueryExceptionHandlerTest.java
vogelsgesang: remind me... why don't we do "backwards compilation"?
Would purely syntactic features (like text blocks) work in backwards compilation mode?

mkaufmann: >remind me... why don't we do "backwards compilation"?

The `--release` flag will only work during java compilation and thus not check if there are incompatibilities in 3PPs. Thus back in december we setup the whole CI pipeline to use a Java 8 JDK so that tests would run with Java 8 to test also 3PP java 8 compatibility.
As the whole CI pipeline is using Java 8 for compile & test there is no "new" frontend there.

Since the move to Gradle I think it should be possible to use a newer JDK for compilation and an Java 8 JDK for testing. @jschneidereit would that work? If yes lets add it to our internal quality uplift work.

@vogelsgesang I don't think there is a quick fix I can do in this PR though

PR #82, opened 2025-07-11
Draft: Simplify waitFor methods and reorganize query v3 classes

jschneidereit:
Consolidate multiple waitFor* methods into a single waitFor method with predicate-based approach for different readiness conditions (general, row-based, and chunk-based).

Rename DataCloudQueryStatus to QueryStatus for consistency with QueryTimeout and move both classes to query.v3 namespace in preparation for extracting into separate library.


jdbc-core/src/main/java/com/salesforce/datacloud/query/v3/HyperClient.java
```diff
@@ -0,0 +1,42 @@
...
+    Iterator<QueryResult> getQueryResult(String queryId, long offset, long rowLimit, long byteLimit, boolean omitSchema)
+            throws DataCloudJDBCException;
+
+    Iterator<QueryResult> getQueryResult(String queryId, long chunkId, boolean omitSchema)
+            throws DataCloudJDBCException;
+
+    Stream<QueryStatus> getQueryStatus(String queryId) throws DataCloudJDBCException;
+
+    void cancel(String queryId) throws DataCloudJDBCException;
+}
```
mkaufmann, line 42: This interface seems completely unnecessary at this point, can we just leave it away or is it important to this change?

I'm reluctant about adding this as I'm not sure that an interface like execute* with just sql and queryTimeout makes sense or whether we wouldn't want to have something like queryparams as parameter there.
Similar for generalizing the client the get query result methods would need a way to get dataspace and other execution context. How all of that is stitched together is unclear.

Thus I hope that we can postpone the introduction of this interface until we have a design for the low level API

jdbc-core/src/main/java/com/salesforce/datacloud/query/v3/QueryStatus.java
```diff
@@ -32,8 +32,8 @@
  * </ul>
  */
 @Value
-@Unstable
-public class DataCloudQueryStatus {
+@Unstable // TODO: it might be time to remove this
```
mkaufmann, line 35: Please add comments (like those from the proto) that describe the different fields.

mkaufmann: The `isResultProduced` public methods feel like it shouldn't be used. Folks should rather use `allResultsProduced` or `isExecutionFinished`. Let's remove it

jdbc-core/src/main/java/com/salesforce/datacloud/query/v3/QueryStatus.java
```diff
@@ -77,17 +77,16 @@ public boolean isExecutionFinished() {
         return completionStatus == CompletionStatus.FINISHED;
     }
 
-    public static Optional<DataCloudQueryStatus> of(QueryInfo queryInfo) {
-        return Optional.ofNullable(queryInfo).map(QueryInfo::getQueryStatus).map(DataCloudQueryStatus::of);
+    public static Optional<QueryStatus> of(QueryInfo queryInfo) {
+        return Optional.ofNullable(queryInfo).map(QueryInfo::getQueryStatus).map(QueryStatus::of);
```
mkaufmann, line 81: Do we need to provide an `Optional<QueryStatus>` interface? Parsing of `QueryInfo` I would keep outside of this class

jdbc-core/src/main/java/com/salesforce/datacloud/query/v3/QueryStatus.java
```diff
@@ -99,4 +98,38 @@ private static CompletionStatus of(QueryStatus.CompletionStatus completionStatus
...
+    /**
+     * Provides a set of suggested predicates for determining the status of a query.
+     */
+    public static class Predicates {
+        /**
+         * Simply get the first query status that the server replies with.
+         */
+        public static Predicate<QueryStatus> first() {
+            return status -> true;
+        }
```
mkaufmann, line 111: For discoverability would this make sense to have as a separate class called `QueryStatusPredicates`?

mkaufmann, line 111: Not sure if we need to provide this as an prebuilt abstraction

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -222,80 +222,37 @@ public DataCloudResultSet getRowBasedResultSet(String queryId, long offset, long
...
         return getChunkBasedResultSet(queryId, chunkId, 1);
     }
 
     /**
-     * Checks if a given row range is available for a query.
-     * This method will wait until the row range specified by the other params is available in the given timeout.
-     *
+     * Checks if a given predicate is satisfied by the status of the specified query.
+     * This method will poll the query status and exit immediately when the predicate is satisfied.
+     * If the predicate is never satisfied, the method will continue trying until the provided timeout is reached.
```
mkaufmann, line 243: Waiting until timeout if the predicate is never satisfied sounds like the wrong behavior. If the query is finished and nothing changes anymore it should finish earlied

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -222,80 +222,37 @@ public DataCloudResultSet getRowBasedResultSet(String queryId, long offset, long
...
      * @param queryId The identifier of the query to check
-     * @param offset The starting row offset.
-     * @param limit The quantity of rows relative to the offset to wait for
      * @param timeout The duration to wait for the engine have results produced.
-     * @param allowLessThan Whether to return early when the available rows is less than {@code offset + limit}
-     * @return The first status where the rows available meet the constraints defined by the parameters or the last status the server replied with.
+     * @param predicate The condition to check against the query status,
+     *                  you can use {@link QueryStatus#isExecutionFinished()}, {@link QueryStatus#allResultsProduced()}, and {@link QueryStatus#isResultProduced()},
+     *                  as well as the methods on {@link QueryStatus.Predicates} for more complex predicates involving pages of rows or chunks
+     * @return The first status that satisfies the predicate, or the last {@link QueryStatus} received before timeout.
```
mkaufmann, line 250: Should we rather throw an exception if the query status condition couldn't be met? That way it would be super explicit that the method will only return if the condition was met

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudResultSet.java
```diff
@@ -15,13 +15,8 @@
...
 
-import com.salesforce.datacloud.jdbc.exception.DataCloudJDBCException;
-import com.salesforce.datacloud.query.v3.DataCloudQueryStatus;
 import java.sql.ResultSet;
-import java.util.stream.Stream;
 
 public interface DataCloudResultSet extends ResultSet {
     String getQueryId();
-
-    Stream<DataCloudQueryStatus> getQueryStatus() throws DataCloudJDBCException;
```
mkaufmann, line 26: This change doesn't seem to be described in the commit message?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/HyperGrpcClientExecutor.java
```diff
@@ -90,12 +90,13 @@ public HyperGrpcClientExecutor withQueryParams(QueryParam additionalQueryParams)
         return this;
     }
 
-    public Iterator<ExecuteQueryResponse> executeQuery(String sql, QueryTimeout queryTimeout) throws SQLException {
+    public Iterator<ExecuteQueryResponse> executeQuery(String sql, QueryTimeout queryTimeout)
+            throws DataCloudJDBCException {
```
mkaufmann, line 94: Why can't this continue to be a SQLException? We currently use `DataCloudJDBCException` in a lot of places where it does provide no value over a normal SQLException. Wouldn't we want to rather change those places to throw a normal SQLException?

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/partial/DataCloudQueryPollingTest.java
```diff
@@ -40,41 +38,17 @@ class DataCloudQueryPollingTest {
...
-                    .isTrue();
-
-            Assertions.assertThatThrownBy(() -> connection.waitForRowsAvailable(
-                            statement.getQueryId(), 100, 10, Duration.ofSeconds(1), true))
-                    .hasMessageContaining("Timed out waiting for new items to be available")
-                    .isInstanceOf(DataCloudJDBCException.class);
+            val result = connection.waitFor(
+                    statement.getQueryId(), Duration.ofMillis(100), QueryStatus.Predicates.rowsAvailable(100, 10));
+            assertThat(result.getRowCount()).isEqualTo(109);
+            assertThat(result.allResultsProduced()).isTrue();
```
mkaufmann, line 51: Revisit these test cases when we have a decision on whether waitFor should throw

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/examples/RowBasedPaginationTest.java
```diff
@@ -86,12 +86,14 @@ public void testRowBasedPagination() throws SQLException {
...
-            DataCloudQueryStatus status = conn.waitForRowsAvailable(queryId, currentOffset, pageSize, timeout, true);
+            QueryStatus status =
+                    conn.waitFor(queryId, timeout, QueryStatus.Predicates.rowsAvailable(currentOffset, pageSize));
 
             while (true) {
                 final boolean shouldCheck = !status.allResultsProduced() && currentOffset >= status.getRowCount();
                 if (shouldCheck) {
-                    status = conn.waitForRowsAvailable(queryId, currentOffset, pageSize, timeout, true);
+                    status = conn.waitFor(
+                            queryId, timeout, QueryStatus.Predicates.rowsAvailable(currentOffset, pageSize));
```
mkaufmann, line 96: The example code doesn't handle timeouts correctly. In case of a timeout it would now just go to `getRowBasedResultSet` requesting an invalid page range

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/examples/ResultScanTest.java
```diff
@@ -59,7 +59,7 @@ void testResultScanWithWait() {
 
         try (val conn = DataCloudConnection.of(channelBuilder, properties);
                 val stmt = conn.createStatement()) {
-            conn.waitForQueryStatus(queryId, Duration.ofDays(1), DataCloudQueryStatus::isExecutionFinished);
+            conn.waitFor(queryId, Duration.ofDays(1), QueryStatus::isExecutionFinished);
```
mkaufmann, line 62: This example doesn't handle timeouts correctly

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/examples/ChunkBasedPaginationTest.java
```diff
@@ -60,12 +60,11 @@ public void testChunkBasedPagination() throws SQLException {
...
 
         int prev = 1;
-        DataCloudQueryStatus status = null;
+        QueryStatus status = null;
         while (true) {
             try (final DataCloudConnection conn = DataCloudConnection.of(channelBuilder, properties)) {
                 if (status == null || !status.allResultsProduced()) {
-                    status = conn.waitForChunksAvailable(
-                            queryId, offset.get(), 1, timeout, false); // false because we're waiting for the next chunk
+                    status = conn.waitFor(queryId, timeout, QueryStatus.Predicates.chunksAvailable(offset.get(), 1));
```
mkaufmann, line 67: This example doesn't handle timeouts correctly

jdbc-core/src/test/java/com/salesforce/datacloud/query/v3/QueryStatusTest.java
```diff
@@ -0,0 +1,175 @@
...
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Named;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import salesforce.cdp.hyperdb.v1.QueryInfo;
+
+class QueryStatusTest {
+    private static QueryInfo random(Consumer<salesforce.cdp.hyperdb.v1.QueryStatus.Builder> update) {
```
mkaufmann, line 38: What is random about this?

jdbc-core/src/main/java/com/salesforce/datacloud/query/v3/QueryStatus.java
jschneidereit: Because it's a static inner class it feels pretty discoverable, it's also referenced in the javadoc for the top-level `waitFor` which I thought you or Adrian had said was good for discoverability 🤔

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/examples/RowBasedPaginationTest.java
```diff
@@ -78,29 +78,33 @@ public void testRowBasedPagination() throws SQLException {
...
 
-                final boolean readAllRows = status.allResultsProduced() && currentOffset >= status.getRowCount();
+                final boolean readAllRows = status.allResultsProduced() && currentOffset.get() >= status.getRowCount();
                 if (readAllRows) {
                     break;
                 }
 
-                final DataCloudResultSet rs =
-                        conn.getRowBasedResultSet(queryId, currentOffset, pageSize, RowBased.Mode.SINGLE_RPC);
+                final DataCloudResultSet rs = conn.getRowBasedResultSet(queryId, currentOffset.get(), pageSize);
```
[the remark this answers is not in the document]
mkaufmann: Actually I think the best approach is to just simplify the interface to
```
conn.getRowBasedResultSet(queryId, offset)
```
and remove the variant with limit. The behavior would be exactly like
```
... create full result set
rs.absolute(offset)
```
and it would just allow to provide a cheap variant to get this behavior. The user can then figure out how to track the number of consumed rows / page limit themselves and it should be a problem. And by defining that it behaves like full result set + absolute the `getRow()` semantics would also be clarified that it should be the row id within query wide result set

jdbc-core/src/main/java/com/salesforce/datacloud/query/v3/QueryStatus.java
jschneidereit: This can be removed once the `Stream` heavy `QueryStatusListener` implementations are replaced

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/examples/RowBasedPaginationTest.java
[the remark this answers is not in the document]
jschneidereit: Couldn't we just avoid this by saying that page bounds should be calculated based on a QueryStatus? We could have an overload like: `getRowBasedResultSet(queryId, offset, status)` and calculate the limit internally

jschneidereit:
Replaced by #92

PR #85, opened 2025-07-25
Fix: replace OOM succeptible Stream API based Adaptive query client with Finite State Machine implementation

jschneidereit:
This pull request resolves an OutOfMemoryError described below by replacing the Adaptive query client that was formerly based on the Java Stream API with a one based on a Finite State Machine. 

Additionally, we reduce the unnecessary size of the queries in our integration tests (originally this was to enforce multiple chunks per query) by using the correct configuration in our hyper.yaml file, therefore reducing the overall test time from a couple minutes down to ~20 seconds.

## Original proposal

This documents the proposed solution to address memory consumption issues in the current `AdaptiveQueryStatusListener` implementation. The existing code uses extensive Stream API constructs (flatMap, filter, map) in conjunction with Spliterator-based iteration which causes `OutOfMemoryError: Java heap space` when processing large result sets. This may have been either introduced or exacerbated when we migrated from Java 11 to Java 8 since this overload of Stream::iterate was no longer available to us:
[`Stream<T> iterate​(T seed, Predicate<? super T> hasNext, UnaryOperator<T> next)`](https://docs.oracle.com/javase/9/docs/api/java/util/stream/Stream.html#iterate-T-java.util.function.Predicate-java.util.function.UnaryOperator-).

The updated `SubmitQueryAndConsumeResultsTest` demonstrates the current issue and throws `OutOfMemoryError: Java heap space` with the existing streaming-based implementation. This FSM design will guide the implementation of a memory-efficient solution that maintains Iterator<QueryResult> interface while using state-driven API calls under the hood.

Solution: The proposed FSM eliminates Stream API constructs entirely, making targeted `GetQueryResult` and `GetQueryInfo` gRPC calls based on state transitions instead of relying on Stream transformations. This should provide true streaming semantics with minimal memory footprint while maintaining (even improving) the existing Adaptive transfer mode benefits  maintaining the Adaptive transfer mode benefits of faster time-to-first-result than pure async mode while avoiding timeout risks of sync mode.

We expect this solution to not only be more efficient than the existing implementation, but also to have better performance characteristics since the current implementation currently calls `waitForQueryStatus` wait for the async portion (tail) of the result set as opposed to the FSM which will acquire chunks as they become available, see [AdaptiveQueryStatusListener#tail](jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/listener/AdaptiveQueryStatusListener.java).

References:
- [Why the tryAdvance of stream.spliterator() may accumulate items into a buffer?](https://stackoverflow.com/questions/47036993/why-the-tryadvance-of-stream-spliterator-may-accumulate-items-into-a-buffer)
- [Why filter() after flatMap() is "not completely" lazy in Java streams?](https://stackoverflow.com/questions/29229373/why-filter-after-flatmap-is-not-completely-lazy-in-java-streams)

### Adaptive Query Enhancements
* Added documentation for the adaptive query pattern in `docs/adaptive-fsm.md`, including a detailed state diagram explaining the flow of query execution and result fetching phases.

### Refactoring Query Execution and Result Handling
* Replaced `AsyncQueryStatusListener` and `AdaptiveQueryStatusListener` with `AsyncQueryResultHandle` and `AdaptiveQueryResultIterator` for better encapsulation of query state and results. Updated methods like `executeAdaptiveQuery` and `executeAsyncQuery` in `DataCloudStatement` to use these new classes. [[1]](diffhunk://#diff-03b48f2d320d22350bf328a0ab3cb10ecdb3585ab5dd678f9e79142e8ec5ee46L115-R135) [[2]](diffhunk://#diff-e822acfd0a73654b02524fb194aead2740b5dd31d4168b9c6b0b99e16e60bfaaL113-R120)
* Updated `StreamingResultSet` to use `QueryResultIterator` instead of relying on a client and listener, simplifying result set creation and management.

### Simplifying Method Naming
* Renamed `getQueryExecutor` to `getQueryClient` in `DataCloudStatement` and `DataCloudPreparedStatement` for better alignment with its purpose. [[1]](diffhunk://#diff-03b48f2d320d22350bf328a0ab3cb10ecdb3585ab5dd678f9e79142e8ec5ee46L69-R87) [[2]](diffhunk://#diff-e822acfd0a73654b02524fb194aead2740b5dd31d4168b9c6b0b99e16e60bfaaL98-R98)

### Code Cleanup and Consistency
* Removed unused imports and methods, such as `getQueryStatus` from `DataCloudResultSet` and redundant listener imports from `DataCloudStatement`. [[1]](diffhunk://#diff-8e3b14540a19a6779d7571e8d4f9d9fbc4927ff4aa46952aeb89ec7936243692L18-L26) [[2]](diffhunk://#diff-03b48f2d320d22350bf328a0ab3cb10ecdb3585ab5dd678f9e79142e8ec5ee46L18-R28)
* Updated parameter names for clarity, such as renaming `timeout` to `waitTimeout` in `waitForResultsProduced` in `DataCloudConnection`.

### Minor Fixes and Improvements
* Replaced direct exception instantiation with a factory method (`createException`) for better consistency in error handling.
* Adjusted memory allocation logic in `StreamingResultSet` to streamline resource management.


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/fsm/AdaptiveQueryResultIterator.java
```diff
@@ -0,0 +1,267 @@
...
+                if (context.executeQueryResponses != null && context.executeQueryResponses.hasNext()) {
+                    val response = context.executeQueryResponses.next();
+
+                    if (response.hasQueryInfo()) {
+                        context.updateQueryContext(response.getQueryInfo());
+                    }
+
+                    if (response.hasQueryResult()) {
+                        context.setQueryResult(response.getQueryResult());
+                    }
```
mkaufmann, line 128: These should be if ... elif ... and then an else where we throw an exception if the message is not marked as optional

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -60,14 +60,20 @@ private StreamingResultSet(
...
+
+    public static StreamingResultSet of(QueryResultIterator iterator) throws DataCloudJDBCException {
+        return of(iterator, iterator::getQueryId, iterator::getQueryStatus);
     }
 
-    @SneakyThrows
     public static StreamingResultSet of(
-            String queryId, HyperGrpcClientExecutor client, Iterator<QueryResult> iterator) {
+            Iterator<QueryResult> iterator,
+            ThrowingJdbcSupplier<String> getQueryId,
```
mkaufmann, line 74: Where would this actually be throwing for queryId? I didn't find a place where that would happen. Looks like it can just stay String

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -60,14 +60,20 @@ private StreamingResultSet(
...
+            ThrowingJdbcSupplier<DataCloudQueryStatus> getQueryStatus)
```
mkaufmann, line 75: Why not just `DataCloudQueryStatus` for an initial status and if it is null / if not finished yet and we need to update the ResultSet can just fetch the status itself using the queryId. That would make the logic more encapsulated instead of having an arbitrary supplier to reason around in the result set

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudResultSet.java
```diff
@@ -18,10 +18,15 @@
 import com.salesforce.datacloud.jdbc.exception.DataCloudJDBCException;
 import com.salesforce.datacloud.query.v3.DataCloudQueryStatus;
 import java.sql.ResultSet;
-import java.util.stream.Stream;
+import java.time.Duration;
+import java.util.function.Predicate;
 
 public interface DataCloudResultSet extends ResultSet {
-    String getQueryId();
+    String getQueryId() throws DataCloudJDBCException;
```
mkaufmann, line 25: Without the supplier design this could also be not throwing

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -261,9 +261,18 @@ public ResultSet getResultSet() throws SQLException {
...
-            resultSet = listener.generateResultSet();
+            if (iterator instanceof AsyncQueryResultIterator) {
+                val status = iterator.getQueryStatus();
+                log.warn(
+                        "Prefer acquiring async result sets from the helper methods on DataCloudConnection, will use chunk based result set from status={}",
+                        status);
+                resultSet = connection.getChunkBasedResultSet(status.getQueryId(), 0, status.getChunkCount());
+            } else {
+                resultSet = StreamingResultSet.of(iterator);
+            }
```
mkaufmann, line 272: This logic looks wrong to me. `getQueryStatus` doesn't guarantee that the query has results produced so this might produce an arbitrary fraction of results from the overall result.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/HyperGrpcClientExecutor.java
```diff
@@ -219,25 +219,24 @@ private QueryInfoParam getQueryInfoParam(String queryId) {
...
-                    val request = getQueryParams(sql, builder);
-                    // We set the deadline based off the query timeout here as the server-side doesn't properly enforce
-                    // the query timeout during the initial compilation phase. By setting the deadline, we can ensure
-                    // that the query timeout is enforced also when the server hangs during compilation.
-                    val remainingDuration = queryTimeout.getLocalDeadline().getRemaining();
-                    return stub.withDeadlineAfter(remainingDuration.toMillis(), TimeUnit.MILLISECONDS)
-                            .executeQuery(request);
-                },
+                () -> stub.withDeadlineAfter(remainingDuration.toMillis(), TimeUnit.MILLISECONDS)
+                        .executeQuery(request),
```
mkaufmann, line 228: Please add comment agai

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/fsm/AdaptiveQueryResultIterator.java
```diff
@@ -0,0 +1,267 @@
...
+        void updateQueryContext(QueryInfo info) {
+            DataCloudQueryStatus.of(info).ifPresent(status::set);
+        }
+
+        String getQueryId() {
+            val current = status.get();
+            return current != null ? current.getQueryId() : null;
+        }
+
+        boolean isQueryComplete() {
```
mkaufmann, line 219:
```suggestion
        boolean allResultsProduced() {
```

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/StreamingResultSetTest.java
```diff
@@ -40,7 +40,7 @@ public static String query(String arg) {
     }
 
     private static final Properties none = new Properties();
-    private static final int large = 10 * 1024 * 1024;
+    private static final int large = 64;
```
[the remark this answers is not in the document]
jschneidereit: I'm going to revert this change since it this depends on the `result_target_chunk_size * arrow_write_buffer_initial_tuple_limit` config

buildSrc/src/main/kotlin/java-base-conventions.gradle.kts
```diff
@@ -25,6 +25,4 @@ tasks.withType<Test>().configureEach {
         showStandardStreams = true
         showStackTraces = true
     }
-
-    jvmArgs("-Xmx2g", "-Xms512m")
```
jschneidereit, line 29: This appears to have been hiding the OOM issue

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
jschneidereit: Thanks for this comment, it helped me arrive at a better abstraction I think. Async -> Handle and Adaptive -> Iterator. The handle is intended to be used with waitFor/get{Row|Chunk}Based while the iterator can directly be fed to a StreamingResultSet

PR #92, opened 2025-08-08
BREAKING: replace waitFor* methods with predicate-based waitFor and remove RowBased.Mode

jschneidereit:
Replace Java Stream API with FSM for query status polling to fix lazy evaluation issues.
Remove count-based waitFor methods in favor of predicate-based API for simplified usage.

API Changes:
- Remove waitForChunksAvailable(), waitForRowsAvailable(), waitForResultsProduced()
- Remove getQueryStatus() Stream-based method
- Replace waitForQueryStatus() with waitFor() accepting Predicate<QueryStatus>
- Add waitFor() overload with infinite timeout (no Duration parameter)
- Remove RowBased.Mode enum and interface, keep only full-range behavior
- Rename DataCloudQueryStatus to QueryStatus

Exception Behavior:
- waitFor() now throws DataCloudJDBCException when predicate fails on finished execution
- waitFor() throws DataCloudJDBCException on timeout exceeded
- Improved error messages with better context

Breaking Changes: Consumers must migrate from count-based waitFor* methods to
predicate-based waitFor() using QueryStatus boolean methods like:
  waitFor(queryId, s -> s.allResultsProduced() || s.getRowCount() >= targetRows)


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -238,64 +271,25 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
...
-     * This method will wait until the server responds with a satisfactory status or the timeout is reached.
+     * This method will wait until the server responds with a satisfactory status, the query is finished, or the timeout is reached.
      * @param queryId The identifier of the query to check
      * @param waitTimeout The duration to wait for the engine have results produced.
-     * @return The first satisfactory status or the last {@link DataCloudQueryStatus} the server replied with.
+     * @return The first satisfactory status or the last {@link QueryStatus} the server replied with.
      */
-    public DataCloudQueryStatus waitForQueryStatus(
-            String queryId, Duration waitTimeout, Predicate<DataCloudQueryStatus> predicate)
+    public QueryStatus waitFor(String queryId, Duration waitTimeout, Predicate<QueryStatus> predicate)
```
mkaufmann, line 289: @vogelsgesang can you check if this matches your understanding on the discussion for `waitFor` behavior?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -238,64 +271,25 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
...
     }
 
     /**
      * Checks if a given predicate is satisfied by the status of the query.
-     * This method will wait until the server responds with a satisfactory status or the timeout is reached.
+     * This method will wait until the server responds with a satisfactory status, the query is finished, or the timeout is reached.
      * @param queryId The identifier of the query to check
      * @param waitTimeout The duration to wait for the engine have results produced.
-     * @return The first satisfactory status or the last {@link DataCloudQueryStatus} the server replied with.
+     * @return The first satisfactory status or the last {@link QueryStatus} the server replied with.
```
mkaufmann, line 287: Document on when this can throw. Is it only if the query fails?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -238,64 +271,25 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
...
-    }
-
-    /**
-     * Checks if all the query's results are ready, the row count and chunk count will be stable.
-     * @param queryId The identifier of the query to check
-     * @param waitTimeout The duration to wait for the engine have results produced.
-     * @return The first status where {@link DataCloudQueryStatus#allResultsProduced()} or the last status the server replied with.
+     * Checks if a given predicate is satisfied by the status of the query.
+     * This method will wait until the server responds with a satisfactory status or the query is finished.
+     * @see #waitFor(String, Duration, Predicate)
```
mkaufmann, line 276: Document on when this can throw. Is it only if the query fails?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -238,64 +271,25 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
...
-        return executor.waitForChunksAvailable(queryId, offset, limit, Deadline.of(waitTimeout), allowLessThan);
-    }
-
-    /**
-     * Checks if all the query's results are ready, the row count and chunk count will be stable.
-     * @param queryId The identifier of the query to check
-     * @param waitTimeout The duration to wait for the engine have results produced.
-     * @return The first status where {@link DataCloudQueryStatus#allResultsProduced()} or the last status the server replied with.
+     * Checks if a given predicate is satisfied by the status of the query.
+     * This method will wait until the server responds with a satisfactory status or the query is finished.
```
mkaufmann, line 275: Add documentation on how long this will wait

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/examples/ResultScanTest.java
```diff
@@ -61,7 +61,7 @@ void testResultScanWithWait() {
                 val stmt = conn.createStatement()) {
             // We don't have any separate query timeouts here, as Spark already has a global job timeout, anyway.
             // TODO (W-18851398): Set the timeout to infinite, as soon as `waitForQueryStatus` accepts it.abstract
-            conn.waitForQueryStatus(queryId, Duration.ofDays(1), DataCloudQueryStatus::isExecutionFinished);
+            conn.waitFor(queryId, Duration.ofDays(1), QueryStatus::isExecutionFinished);
```
mkaufmann, line 64: This is showing wrong usage of `waitFor` as new behavior is that it returns even on timeout / if pred is not met. Thus it's not guaranteed that we are allowed to proceed with the execute query. Would adjust to capture the returned status and throw an exception if the query is not finished

jschneidereit: The new behavior is that we throw if timeout is hit and predicate is not satisfied, I've updated the javadocs for waitFor to better describe that

PR #100, opened 2025-08-12
ci: use gradle wrapper in ci per https://github.com/gradle/actions

jschneidereit:


.github/workflows/reusable-build-publish.yml
```diff
@@ -42,14 +39,11 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '17'
-      - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@v4
-        with:
-          gradle-version: ${{ env.GRADLE_VERSION }}
+      - uses: gradle/actions/setup-gradle@v4
```
vogelsgesang, line 42: do we still need this action at all, if we are calling `gradlew` anyway?

Here and all other uses of `gradle/actions/setup-gradle@v4`

carsheng (approved):

.github/workflows/reusable-build-publish.yml
jschneidereit: I'll remove the first one, but the main build sets up build scanning etc and we should keep that

jschneidereit: I tried removing them as an experiment and simple things like `./gradlew -q printHyperApiVersion` took 3x as long

PR #107, opened 2025-08-21
fix: Validate JDBC settings properties and throw user errors

reblem:
- Added `PropertyValidator` to reject unknown JDBC properties and allow querySetting.*/grpc.* prefixes.
- Enforced querySetting. prefix for session settings (e.g., use `querySetting.lc_time`, `querySetting.time_zone`).
- Validated properties early in DataCloudJDBCDriver and DataCloudConnection.
- Updated tests to match stricter validation and existing polling messages.
- Goals achieved: clearer user errors, prevents silent misconfigurations; existing querySetting.* users unaffected

### DIFF
--- a/jdbc-util/src/main/java/com/salesforce/datacloud/jdbc/util/PropertyValidator.java
+++ b/jdbc-util/src/main/java/com/salesforce/datacloud/jdbc/util/PropertyValidator.java
@@ -65,9 +65,6 @@ public class PropertyValidator {
             return "";
         }
         String normalized = rawKey.trim().toLowerCase().replace('-', '_');
-        if ("timezone".equals(normalized)) {
-            return "time_zone";
-        }
         return normalized;
     }
 
--- a/jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/PropertiesTest.java
+++ b/jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/PropertiesTest.java
@@ -129,13 +129,12 @@ class PropertiesTest extends HyperGrpcTestBase {
 
     @Test
     void testUnprefixedTimezoneRaisesUserError() {
-        // Users sometimes set TIMEZONE/time_zone directly without the required querySetting. prefix
-        for (String key : ImmutableList.of("TIMEZONE", "time_zone")) {
-            Properties properties = new Properties();
-            properties.setProperty(key, "Asia/Tokyo");
-            val exception = assertThrows(DataCloudJDBCException.class, () -> ConnectionProperties.of(properties));
-            assertThat(exception.getMessage()).contains("Use 'querySetting.time_zone'");
-        }
+        // Users sometimes set time_zone directly without the required querySetting. prefix
+        // Only time_zone should trigger the targeted hint here.
+        Properties properties = new Properties();
+        properties.setProperty("time_zone", "Asia/Tokyo");
+        val exception = assertThrows(DataCloudJDBCException.class, () -> ConnectionProperties.of(properties));
+        assertThat(exception.getMessage()).contains("Use 'querySetting.time_zone'");
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


jdbc-util/src/main/java/com/salesforce/datacloud/jdbc/util/PropertyValidator.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Validates incoming JDBC properties and raises user errors for unknown properties.
+ *
+ * This does not validate presence of required keys (handled elsewhere), and it
+ * deliberately allows any subkey under the {@code querySetting.} namespace to
+ * defer detailed validation to the server.
+ */
+public final class PropertyValidator {
```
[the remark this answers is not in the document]
reblem: Good point, will update this class to use that.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -113,8 +114,9 @@ public static DataCloudConnection of(@NonNull HyperGrpcStubProvider stubProvider
      */
     public static DataCloudConnection of(@NonNull ManagedChannelBuilder<?> builder, @NonNull Properties properties)
             throws DataCloudJDBCException {
-        val stubProvider = new JdbcDriverStubProvider(DataCloudJdbcManagedChannel.of(builder, properties), true);
-        return of(stubProvider, properties);
+        val validated = validate(properties);
```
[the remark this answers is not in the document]
KaviarasuSakthivadivel: Also, `Properties validate(Properties properties)` doesn't change the properties value right? may be we can keep it void.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StatementProperties.java
```diff
@@ -62,6 +62,24 @@ public class StatementProperties {
...
+            String normalized = rawKey.toLowerCase();
+            // Common session setting synonyms users might try to pass without prefix
+            if ("time_zone".equals(normalized) || "timezone".equals(normalized)) {
+                throw new DataCloudJDBCException("Invalid property '" + rawKey
+                        + "'. Use 'querySetting.time_zone' to set the session time zone.");
+            }
+            if ("lc_time".equals(normalized)) {
+                throw new DataCloudJDBCException(
+                        "Invalid property '" + rawKey + "'. Use 'querySetting.lc_time' to set the session locale.");
+            }
```
jschneidereit, line 80: I like this, is there a set of settings for hyper? We might consider making this generic for anything in that set:

```java
static final Set<String> normalizedCommonHyperSettings = ...

if (commonHyperSettings.contains(normalized)) {
       throw new DataCloudJDBCException(
                        "Invalid property '" + rawKey + "'. Use 'querySetting." + normalized + "' to set the query setting.");
```

jschneidereit: I'd probably move this to `PropertyValidator` under like `PropertyValidator.validateQuerySettings` since you do have that set there

reblem: Added common Hyper settings. Is there a more complete list?

jdbc-util/src/main/java/com/salesforce/datacloud/jdbc/util/PropertyValidator.java
```diff
@@ -0,0 +1,112 @@
...
+     * "querySetting." prefix. We proactively validate these to provide a helpful error message.
+     */
+    private static final Set<String> COMMON_HYPER_SETTINGS = ImmutableSet.of("time_zone", "lc_time");
+
+    private static String canonicalizeSettingName(String rawKey) {
+        if (rawKey == null) {
+            return "";
+        }
+        String normalized = rawKey.trim().toLowerCase().replace('-', '_');
+        if ("timezone".equals(normalized)) {
```
jschneidereit, line 68: I think this might be too much, I'm happy with the `-` -> `_` because it makes [url settings](https://github.com/forcedotcom/datacloud-jdbc/blob/main/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnectionString.java#L109) not need to be url encoded, but we've been working towards just letting the server through on things that it doesn't know about as well as being explicit and throwing instead of "helping" the user like this so that they know when they're doing something wrong. 

In short, let's remove `timezone` -> `time_zone` key conversion

carsheng (approved):

jschneidereit (approved):

PR #109, opened 2025-08-29
feat: adds DataCloudConnection::getSchemaForQueryId

jschneidereit:
When a result set returns no rows it can be difficult to work with when using async query methods. Now when you detect that your result set has no rows by using `DataCloudConnection::waitFor` you can use `DataCloudConnection::getSchemaForQueryId` which will return a result set with a schema wired up. 

A logical extension from this would be `getResultSetForQueryId` which could just be a refactor to `AdaptiveQueryResultIterator` to add an overload that accepts a query id and either acquires the results or an empty result set with a schema depending on what it detects about that query.


himanshu-mishra-salesforce:
Continuing development in this PR - https://github.com/forcedotcom/datacloud-jdbc/pull/113

PR #112, opened 2025-09-12
feat: Revamp connection string and property handling

vogelsgesang:
This commit revamps the upper layers (driver, property parsing, URL parsing, ...; everything above `DataCloudConnection`).

It achieves 5 high-level goals:
1. The Spark datasource now connects with Hyper via Java's official DriverManager. It now accepts an JDBC-URL and understands all properties also supported by our normal JDBC driver. All options (workload, ...) are now available also in the Spark data source.
2. The jdbc-core package now exposes its own self-contained JDBC driver, offering the `jdbc:salesforce-hyper:` protocol. Thereby, we will be able to use the Spark driver either directly against a Hyper node using a `jdbc:salesforce-hyper:` URL or against Data Cloud using `jdbc:salesforce-datacloud:` URLs. Note that `jdbc:salesforce-hyper:` is available in the jdbc-core package, which makes it possible to use the Spark datasource without pulling in all the additional dependencies pulled in by the jdbc package.
3. Better URL parsing, using the official Java classes instead of ad-hoc string manipulation. This also means, that URLs no longer use ; to separate paramters. Instead, they now use the usual ?.
4. Strict property parsing. Fail-early, fail-hard, less bugs. Unknown properties are now rejected. Properties specified both in the URL and in the property bag are now rejected. Specifying both user and userName at the same time is now an error. Negative duration / timeout values are now rejected.
5. Consistent property naming scheme: So far, we did not use a consistent naming scheme. Most settings were camelCase (`queryTimeout`, `clientId`, `clientSecret`, ...). A couple of misnormers were also present (e.g., `external-client-context`). We now use a consistent naming scheme.
6. Typesafe configuration using the *Properties classes and the corresponding *Properties.builder() functions in combination with the {Hyper,DataCloud}Datasource. While this gives us a good baseline interface, I expect it to still evolve slightly over time.

Along the way, I had to read and understand quite a bit of code, and refactored large parts of it. I removed quite a few inconsistencies, non-checked error conditions, overly complex class hierarchies:

* The DataCloudJdbcManagedChannel was merged with JdbcDriverStubProvider as the separate abstractions did not lift their weight.
* Simplified the Salesforce-authentication code by:
   1. removing the unnecessary sub-classing of the `*Settings` and the `*Strategy` class hierarchies.
   2. Removing the unnecessarily complex `TokenCache` class hierarchy.
   3. Many utility functions were moved closer to their actual usage.
* The `DEVELOPMENT.md` documentation was improved. Among others, it now documents all supported JDBC properties.
* Couple of deprecated APIs were migrated to up-to-date APIs.

And quite a few of bug fixes / straight-forward cleanups:

* In direct mode, we so far simply ignored all parameters passed in via URL parameters. Now we respect them.
* The HTTP User-Agent is now used for the authentication flow, the dataspaces-related requests and the grpc requests (previously only the HTTP User-Agent was only used for the HTTP flow). Also, the User-Agent is now hard-coded to our driver identification string.
* We now check the provided loginURL against our well-known URLs, not only for key-based authentication
* `RevokeTokenAuthenticationStrategy` was dead code. Now it's gone.
* Fix `InterceptedHyperTestBase` (previously HyperGrpcTestBase) to also use the `HyperServerManager`

### DIFF
--- a/buildSrc/src/main/kotlin/shading.gradle.kts
+++ b/buildSrc/src/main/kotlin/shading.gradle.kts
@@ -52,7 +52,12 @@ abstract class ShadingExtension @Inject constructor(
             
             isReproducibleFileOrder = true
             isPreserveFileTimestamps = false
-            duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+            duplicatesStrategy = DuplicatesStrategy.INCLUDE
+            
+            // Exclude duplicates for non-service files to avoid bloat
+            filesNotMatching("META-INF/services/**") {
+                duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+            }
             
             // JAR naming configuration
             archiveBaseName.set(project.name)
@@ -74,6 +79,8 @@ abstract class ShadingExtension @Inject constructor(
             relocate("org.apache.commons", "$shadeBase.org.apache.commons")
             relocate("org.apache.hc", "$shadeBase.org.apache.hc")
 
+            // Use built-in service file merging with package relocation support
+            // This automatically handles relocated class names in service files!
             mergeServiceFiles {
                 exclude("META-INF/services/java.sql.Driver")
             }


jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DataCloudTokenProvider.java
```diff
@@ -0,0 +1,225 @@
...
+        builder.bodyEntry("client_id", settings.getClientId());
+        builder.bodyEntry("client_secret", settings.getClientSecret());
+
+        switch (settings.getAuthenticationMode()) {
+            case PASSWORD:
+                // https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_username_password_flow.htm
+                builder.bodyEntry("grant_type", "password");
+                builder.bodyEntry("password", settings.getUserName());
+                builder.bodyEntry("username", settings.getPassword());
+                break;
```
praveen2450, line 146:
```suggestion
                builder.bodyEntry("username", settings.getUserName());
                builder.bodyEntry("password", settings.getPassword());
```

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DataCloudTokenProvider.java
```diff
@@ -0,0 +1,225 @@
...
+            val response = FormCommand.post(this.client, command, DataCloudTokenResponse.class);
+            return throwExceptionOnError(
+                    response, "Received an error when exchanging oauth access token for data cloud token");
+        });
+        return DataCloudToken.of(model);
+    }
+
+    public OAuthToken getOAuthToken() throws SQLException {
+        try {
+            // XXX should this also be cached?
```
mkaufmann, line 99: I think it's ok for now without, long term we should implement the token refresh flow (looking at teh code it doesn't seem to be used when using the stock OAuth config

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/SalesforceAuthProperties.java
```diff
@@ -0,0 +1,245 @@
...
+            builder.authenticationMode(AuthenticationMode.REFRESH_TOKEN);
+            builder.refreshToken(takeRequired(props, AUTH_REFRESH_TOKEN));
+            // We still accept an optional userName. This might show up
+            // in the `DatabaseMetadata.getUserName` call.
+            builder.userName(takeOptional(props, AUTH_USER_NAME).orElse(null));
+        } else {
+            throw new DataCloudJDBCException(
+                    "Properties must contain either (userName + password), (userName + privateKey), or refreshToken",
+                    "28000");
+        }
```
mkaufmann, line 161: TIL that we don't support the webserver auth flow :/ (No need to fix)

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ConnectionProperties.java
```diff
@@ -33,46 +29,31 @@ public class ConnectionProperties {
...
-        if (externalClientContextValue != null) {
-            builder.externalClientContext(externalClientContextValue);
-        }
-        String userNameValue = props.getProperty("userName");
-        if (userNameValue != null) {
-            builder.userName(userNameValue);
-        }
-        builder.statementProperties(StatementProperties.of(props));
+        takeOptional(props, "workload").ifPresent(builder::workload);
+        takeOptional(props, "externalClientContext").ifPresent(builder::externalClientContext);
```
mkaufmann, line 55: Please document this breaking change of property name in the commit message (even if it was undocumented but to help with some breadcrumbs)

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -73,81 +77,54 @@ public class DataCloudConnection implements Connection, AutoCloseable {
...
+     * @param dataspacesSupplier a supplier that acquires available dataspace names
+     * @param jdbcUrl The connection URL to use for the connection
      * @return A DataCloudConnection with the given channel and properties
      */
-    public static DataCloudConnection of(@NonNull HyperGrpcStubProvider stubProvider, @NonNull Properties properties)
+    public static DataCloudConnection of(
+            @NonNull HyperGrpcStubProvider stubProvider,
+            @NonNull ConnectionProperties properties,
+            JdbcURL jdbcUrl,
+            @NonNull String userName,
```
mkaufmann, line 97: Why does it have to be non null but allows empty for user name?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StatementProperties.java
```diff
@@ -44,61 +46,35 @@ public class StatementProperties {
...
-
-        // The query timeout property, zero or negative values are interpreted as infinite timeout.
-        // Positive values are interpreted as the number of seconds for the timeout
-        String queryTimeoutStr = props.getProperty("queryTimeout");
-        if (queryTimeoutStr != null) {
-            try {
-                Duration timeout = Duration.ofSeconds(Integer.parseInt(queryTimeoutStr));
-                if (timeout.isZero() || timeout.isNegative()) {
-                    builder.queryTimeout(Duration.ZERO);
-                } else {
```
mkaufmann, line 66: Not supporting negative values is a breaking change, please add it to the commit message

mkaufmann (approved):

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
vogelsgesang: add constructor without user?

PR #113, opened 2025-09-22
feat: adds DataCloudConnection::getSchemaForQueryId

himanshu-mishra-salesforce:
Reference PR - https://github.com/forcedotcom/datacloud-jdbc/pull/109

### WHAT

Through this PR we have:
1.  Added a new method `getSchemaForQueryId` in `DataCloudConnection` which will return `DataCloudResultSet` containing metadata details for clients to infer Schema even if there are no rows in the output result.
2. Started consuming `getSchemaForQueryId` in `HyperResultSource.inferSchema` method to fetch metadata details for the result with zero rows as well which earlier used to fail in such cases.

### WHY

These changes are required to handle queries resulting in zero rows in the output.

### HOW

This is being done by introducing `getSchemaForQueryId` in `DataCloudConnection`, which fetches schema without requiring actual data.

### TESTING

We have added UnitTests covering the scenarios where queries result in zero rows in the output.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -6,7 +6,7 @@
 
 **FluentIterable** from Google Guava is the established reference interface for operating on iterator and stream-like data structures in this codebase. It provides a fluent API for transforming, filtering, and manipulating collections in a functional programming style.
 
-While we also considered the Java Streams API we discourage its use because the `Spliterator` concept together with `tryAdvance` can sneakly lead to OOMs due to materializing full Iterators
+While we also considered the Java Streams API, we discourage its use because the `Spliterator` concept together with `tryAdvance` can sneakily lead to OOMs due to materializing full Iterators.
 
 ## Coverage
 


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -272,6 +272,12 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
         return getChunkBasedResultSet(queryId, chunkId, 1);
     }
 
+    public DataCloudResultSet getSchemaForQueryId(String queryId) throws DataCloudJDBCException {
```
mkaufmann, line 275: I think the old PR had review comments. Can you work those in here as well? For example I think it makes sense that this should be an `ResultSetMetadata` object and likely it should just return a SQLException

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingByteStringChannel.java
```diff
@@ -4,18 +4,68 @@
...
+import salesforce.cdp.hyperdb.v1.QueryInfo;
 import salesforce.cdp.hyperdb.v1.QueryResult;
+import salesforce.cdp.hyperdb.v1.QueryResultPartBinary;
 
-@RequiredArgsConstructor
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
 public class StreamingByteStringChannel implements ReadableByteChannel {
-    private final Iterator<QueryResult> iterator;
+    @AllArgsConstructor
+    private static class ByteStringIterator<T> implements Iterator<Optional<ByteString>> {
```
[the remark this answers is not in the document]
mkaufmann: Hmm looking at this I feel that this class should only handle a single concern: Exposing an `Iterator<ByteString>` as `ReadableByteChannel`. So I would just call it `ByteStringReadableByteChannel`.

The conversion from the V3 specific iterators should happen outside of this class and there would be than different mappers for QueryInfo vs QueryResult. The mixing of concerns has always been confusing me when I came across this class and understand what it's trying to solve

spark-datasource-core/src/test/scala/HyperResultSourceTest.scala
```diff
@@ -211,4 +211,153 @@ class HyperResultSourceTest
...
+
+    // Verify schema is correctly inferred with nullable columns
+    assert(df.schema.fields.length == 5)
+
+    val fields = df.schema.fields
+    assert(
+      fields(0).name == "id" && fields(0).dataType == IntegerType && !fields(
+        0
+      ).nullable
+    )
```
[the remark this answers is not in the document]
himanshu-mishra-salesforce: Updated as mentioned

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/StreamingResultSetTest.java
```diff
@@ -119,6 +120,65 @@ private void assertThatResultSetIsCorrect(DataCloudConnection conn, DataCloudRes
...
+    @Test
+    public void testGetSchemaForQueryIdWithZeroResults() {
+        withStatement(none, (conn, stmt) -> {
+            val sql =
+                    "SELECT s, s::text as s_text, cast(s as numeric(38,18)) as s_numeric FROM generate_series(1,10) s LIMIT 0";
+
+            final String queryId;
+            try (val rs = stmt.executeQuery(sql).unwrap(DataCloudResultSet.class)) {
+                queryId = rs.getQueryId();
+                conn.waitFor(queryId, QueryStatus::allResultsProduced);
```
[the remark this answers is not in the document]
himanshu-mishra-salesforce: Updated as discussed

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -272,6 +276,28 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
...
+            val schemaRoot = reader.getVectorSchemaRoot();
+            val columns = com.salesforce.datacloud.jdbc.util.ArrowUtils.toColumnMetaData(
+                    schemaRoot.getSchema().getFields());
+
+            // Create metadata directly without full ResultSet infrastructure
+            val signature = new Meta.Signature(
+                    columns, null, Collections.emptyList(), Collections.emptyMap(), null, Meta.StatementType.SELECT);
+            return new AvaticaResultSetMetaData(null, null, signature);
+        } catch (Exception ex) {
+            throw new DataCloudJDBCException(ex.getMessage(), ex);
```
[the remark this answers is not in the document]
himanshu-mishra-salesforce: Took a look at it, there is a method `createException` in `QueryExceptionHandler` which was leveraged earlier by `getChunkBasedResultSet`. Will make use of the same here which will keep this consistent to previous behaviour.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -272,6 +276,28 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
...
 
+    public ResultSetMetaData getSchemaForQueryId(String queryId) throws SQLException {
+        val executor = HyperGrpcClientExecutor.forSubmittedQuery(getStub());
+        val infos = executor.getQuerySchema(queryId);
+
+        try {
+            val byteStringIterator = ProtocolMappers.fromQueryInfo(infos);
+            val channel = new ByteStringReadableByteChannel(byteStringIterator);
+            val reader = new org.apache.arrow.vector.ipc.ArrowStreamReader(
+                    channel, new org.apache.arrow.memory.RootAllocator());
```
vogelsgesang, line 287: why use a full `ArrowStreamReader`? Why not directly use [Schema.deserializeMessage](https://arrow.apache.org/docs/18.1/java/reference/org.apache.arrow.vector/org/apache/arrow/vector/types/pojo/Schema.html#deserializeMessage(java.nio.ByteBuffer)) directly?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ByteStringReadableByteChannel.java
```diff
@@ -27,33 +33,29 @@ public int read(ByteBuffer dst) throws IOException {
 
         int totalBytesRead = 0;
 
-        // Continue reading while destination has space AND we have data available
-        while (dst.hasRemaining() && (iterator.hasNext() || (currentBuffer != null && currentBuffer.hasRemaining()))) {
+        while (dst.hasRemaining()
```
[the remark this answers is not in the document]
himanshu-mishra-salesforce: Added back

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ByteStringReadableByteChannel.java
```diff
@@ -28,32 +34,25 @@ public int read(ByteBuffer dst) throws IOException {
...
-                    }
+                if (byteStringIterator.hasNext()) {
+                    ByteString data = byteStringIterator.next();
+                    currentBuffer = data.asReadOnlyByteBuffer();
                 } else {
-                    // This was a non result data query result message like a query info message.
-                    // We ignore that message here and will just try to fetch the next message in
-                    // the next loop iteration.
-                    continue;
+                    break; // No more data
```
himanshu-mishra-salesforce, line 44: @vogelsgesang and @mkaufmann  - Please check this `break` statement, I didn't get what is meant in the comments above, can you please confirm we need `continue` or `break` here, thanks.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ProtocolMappers.java
```diff
@@ -0,0 +1,57 @@
...
+                return queryInfos.hasNext();
+            }
+
+            @Override
+            public ByteString next() {
+                QueryInfo info = queryInfos.next();
+                return info.hasBinarySchema() ? info.getBinarySchema().getData() : null;
+            }
+        };
+    }
```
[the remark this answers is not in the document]
himanshu-mishra-salesforce: Updated as directed, please check, thanks.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
himanshu-mishra-salesforce: Updated to leverage `deserializeMessage` method instead.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -272,6 +283,23 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
         return getChunkBasedResultSet(queryId, chunkId, 1);
     }
 
+    public ResultSetMetaData getSchemaForQueryId(String queryId) throws SQLException {
+        val executor = HyperGrpcClientExecutor.forSubmittedQuery(getStub());
+        val infos = executor.getQuerySchema(queryId);
+
+        try {
+            Schema schema = deserializeMessage(getByteBuffer(queryId, infos));
```
praveen2450, line 291: Should we have getByteBuffer in DirectDataCloudConnection.java, in PR#112 we are removing DirectDataCloudConnection.
cc : @vogelsgesang

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ProtocolMappers.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public static Iterator<ByteString> fromQueryInfo(Iterator<QueryInfo> queryInfos) {
+        return Iterators.filter(
+                Iterators.transform(queryInfos, new Function<QueryInfo, ByteString>() {
+                    @Override
+                    public ByteString apply(QueryInfo input) {
+                        return input.hasBinarySchema() ? input.getBinarySchema().getData() : null;
+                    }
+                }),
+                Predicates.notNull());
```
vogelsgesang, line 36: afaik, usage of guava iterators is discouraged already since Java 8. Nowadays, we should use Java's `stream` APIs instead.

I.e., something like

```
    public static Iterator<ByteString> fromQueryInfo(Iterator<QueryInfo> queryInfos) {
        return StreamSupport.stream(
                        Spliterators.spliteratorUnknownSize(queryInfos, 0),
                        /*parallel=*/ false
                )
                .map(input -> input.hasBinarySchema() ? input.getBinarySchema().getData() : null)
                .filter(Objects::nonNull)
                .iterator();
    }
```

Do you agree that this is a better way to write it, @mkaufmann? Or did I overlook something?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ByteStringReadableByteChannel.java
```diff
@@ -4,21 +4,27 @@
...
+ * This class has a single responsibility: converting ByteString iterator to byte stream.
+ */
+public class ByteStringReadableByteChannel implements ReadableByteChannel {
+    private final Iterator<ByteString> byteStringIterator;
     private boolean open = true;
     private ByteBuffer currentBuffer = null;
 
+    public ByteStringReadableByteChannel(Iterator<ByteString> byteStringIterator) {
+        this.byteStringIterator = Objects.requireNonNull(byteStringIterator, "ByteStringIterator cannot be null");
+    }
```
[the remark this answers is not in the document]
himanshu-mishra-salesforce: updated as mentioned.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -272,6 +283,23 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
...
+            List<ColumnMetaData> columns = toColumnMetaData(schema.getFields());
+
+            // Create metadata directly without full ResultSet infrastructure
+            Meta.Signature signature = new Meta.Signature(
+                    columns, null, Collections.emptyList(), Collections.emptyMap(), null, Meta.StatementType.SELECT);
+            return new AvaticaResultSetMetaData(null, null, signature);
+        } catch (Exception ex) {
+            throw createException(SCHEMA_FAILURE + queryId, ex);
```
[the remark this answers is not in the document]
himanshu-mishra-salesforce: updated as mentioned.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ByteStringReadableByteChannel.java
```diff
@@ -4,21 +4,27 @@
...
 
-@RequiredArgsConstructor
-public class StreamingByteStringChannel implements ReadableByteChannel {
-    private final Iterator<QueryResult> iterator;
+/**
+ * A ReadableByteChannel that exposes an Iterator<ByteString> as a stream of bytes.
+ * This class has a single responsibility: converting ByteString iterator to byte stream.
+ */
+public class ByteStringReadableByteChannel implements ReadableByteChannel {
+    private final Iterator<ByteString> byteStringIterator;
```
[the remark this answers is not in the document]
himanshu-mishra-salesforce: updated as mentioned.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ProtocolMappers.java
mkaufmann: No, don't use Spliterators they have very bad behavior with regard to surprising buffering behavior. I'd argue lets keep the less magic Guava code.

We previously had OOM bugs due to Spliterator (which wouldn't be able to happen here but generally to keep it simple I'd argue we should not use them in the code base given their non obvious behavior) https://stackoverflow.com/questions/47036993/why-the-tryadvance-of-stream-spliterator-may-accumulate-items-into-a-buffer

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudConnection.java
```diff
@@ -284,19 +285,23 @@ public DataCloudResultSet getChunkBasedResultSet(String queryId, long chunkId) t
...
-        val infos = executor.getQuerySchema(queryId);
+        HyperGrpcClientExecutor executor = HyperGrpcClientExecutor.forSubmittedQuery(getStub());
+        Iterator<QueryInfo> infos = executor.getQuerySchema(queryId);
 
         try {
-            Schema schema = deserializeMessage(getByteBuffer(queryId, infos));
+            Iterator<ByteString> byteStringIterator = ProtocolMappers.fromQueryInfo(infos);
+            if (!byteStringIterator.hasNext()) {
+                throw new DataCloudJDBCException("No schema data available for queryId: " + queryId);
+            }
```
vogelsgesang, line 295: Do we have a guarantee in our interface that the schema will always be the first message returned? If so, we might not even need an `byteStringIterator` here?

flode: No, currently the query status is sent first, the schema is sent after that.

vogelsgesang: ok, then we can't exploit that to simplify the code. The code as written is fine by me, then

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ProtocolMappers.java
mkaufmann: FluentIterable looks fine to me. I couldn't immediately find StackOverflow posts that it invites OOMs. I'd be happy with:
- Let's use FluentIterable in the code
-  Establish it as new reference interface to use for operating on iterator / stream like things and add a section to  DEVELOPMENT.md 
- A short comment in the source code which points to DEVELOPMENT.md

himanshu-mishra-salesforce: Updated to FluentIterable

DEVELOPMENT.md
```diff
@@ -1,3 +1,11 @@
+## Development Guidelines
+
+### Best Practices
+
+#### Working with Streams
+
+**FluentIterable** from Google Guava is the established reference interface for operating on iterator and stream-like data structures in this codebase. It provides a fluent API for transforming, filtering, and manipulating collections in a functional programming style.
```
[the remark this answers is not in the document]
himanshu-mishra-salesforce: Updated as mentioned.

DEVELOPMENT.md
```diff
@@ -1,3 +1,13 @@
+## Development Guidelines
+
+### Best Practices
+
+#### Working with Streams and Iterators
+
+**FluentIterable** from Google Guava is the established reference interface for operating on iterator and stream-like data structures in this codebase. It provides a fluent API for transforming, filtering, and manipulating collections in a functional programming style.
+
+While we also considered the Java Streams API we discourage its use because the `Spliterator` concept together with `tryAdvance` can sneakly lead to OOMs due to materializing full Iterators
```
vogelsgesang, line 9:
```suggestion
While we also considered the Java Streams API, we discourage its use because the `Spliterator` concept together with `tryAdvance` can sneakily lead to OOMs due to materializing full Iterators.
```

vogelsgesang (approved):

PR #115, opened 2025-09-26
fix: resolve DataCloudArray data loss after ResultSet exhaustion

KaviarasuSakthivadivel:
- Refactor DataCloudArray to extract and store data during construction
- Fix JDBC array indexing to use proper 1-based indexing
- Add comprehensive tests for string arrays with NULL values

Fixes data loss issue where ValueVector data was lost after ResultSet lifecycle ended, following PostgreSQL JDBC ArrayImpl pattern.

### DIFF
--- a/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/DataCloudArray.java
+++ b/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/DataCloudArray.java
@@ -12,33 +12,64 @@ import java.sql.Array;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Map;
-import lombok.val;
 import org.apache.arrow.memory.util.LargeMemoryUtil;
 import org.apache.arrow.vector.FieldVector;
-import org.apache.arrow.vector.ValueVector;
 
 public class DataCloudArray implements Array {
 
-    private final FieldVector dataVector;
-    private final long startOffset;
-    private final long valuesCount;
+    private final Object[] data;
+    private final String baseTypeName;
+    private final int baseType;
     protected static final String NOT_SUPPORTED_IN_DATACLOUD_QUERY =
             "Array method is not supported in Data Cloud query";
 
     public DataCloudArray(FieldVector dataVector, long startOffset, long valuesCount) {
-        this.dataVector = dataVector;
-        this.startOffset = startOffset;
-        this.valuesCount = valuesCount;
+        // Extract data immediately during construction (following PostgreSQL JDBC pattern)
+        // This makes the object self-contained and independent of ValueVector lifecycle
+        this.data = extractDataFromVector(dataVector, startOffset, valuesCount);
+        this.baseTypeName = toColumnType(dataVector.getField()).getType().getName();
+        this.baseType = toColumnType(dataVector.getField()).getType().getVendorTypeNumber();
+    }
+
+    /**
+     * Extracts data from ValueVector immediately to avoid lifecycle dependencies.
+     * This follows the same pattern as PostgreSQL JDBC: extract data during construction.
+     */
+    private Object[] extractDataFromVector(FieldVector dataVector, long startOffset, long valuesCount) {
+        // Handle empty array case
+        if (valuesCount <= 0) {
+            return new Object[0];
+        }
+
+        // Validate bounds against actual vector size
+        if (startOffset < 0 || startOffset >= dataVector.getValueCount()) {
+            throw new ArrayIndexOutOfBoundsException(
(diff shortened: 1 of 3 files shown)


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/DataCloudArray.java
```diff
@@ -51,7 +82,7 @@ public Object getArray(Map<String, Class<?>> map) throws SQLException {
         if (map != null) {
             throw new DataCloudJDBCException(NOT_SUPPORTED_IN_DATACLOUD_QUERY, SqlErrorCodes.FEATURE_NOT_SUPPORTED);
         }
-        return getArrayNoBoundCheck(this.dataVector, this.startOffset, this.valuesCount);
+        return this.data.clone();
```
mkaufmann, line 85: This should not be required, if a user modifies the array it's their own faul. Having an additional clone would pessimize performance

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/DataCloudArray.java
```diff
@@ -94,17 +127,15 @@ public void free() {
...
-        if (index < 0 || index + count > this.startOffset + this.valuesCount) {
-            throw new ArrayIndexOutOfBoundsException();
+        // Check against actual stored data size using 1-based JDBC indexing
+        if (index < 1 || index > this.data.length) {
+            throw new ArrayIndexOutOfBoundsException("Index " + index + " is out of bounds for array size "
+                    + this.data.length + " (JDBC arrays use 1-based indexing)");
+        }
+        if (index + count - 1 > this.data.length) {
+            throw new ArrayIndexOutOfBoundsException("Index " + index + " + count " + count + " exceeds array size "
+                    + this.data.length + " (JDBC arrays use 1-based indexing)");
```
mkaufmann, line 138: I think the following formulation should be simpler as it follows the normal 0 based bound check logic
```suggestion
        long dataIndex = index - 1;
        // Check against actual stored data size using 1-based JDBC indexing
        if (dataIndex < 0 || dataIndex >= this.data.length) {
            throw new ArrayIndexOutOfBoundsException("Index " + index + " is out of bounds for array size "
                    + this.data.length + " (JDBC arrays use 1-based indexing)");
        }
        if ((dataIndex + count) >= this.data.length) {
            throw new ArrayIndexOutOfBoundsException("Index " + index + " + count " + count + " exceeds array size "
                    + this.data.length + " (JDBC arrays use 1-based indexing)");
```

jdbc-core/src/test/resources/simplelogger.properties
```diff
@@ -1,2 +1,2 @@
 org.slf4j.simpleLogger.logFile=System.out
-org.slf4j.simpleLogger.defaultLogLevel=warn
+org.slf4j.simpleLogger.defaultLogLevel=info
```
mkaufmann, line 2: Nit: Remove

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/accessor/impl/DataCloudArrayTest.java
```diff
@@ -108,6 +113,121 @@ void testShouldThrowIfGetArrayHasIncorrectIndex() {
...
+
+    @SneakyThrows
+    @Test
+    void testGetArrayWithZeroCount() {
+        // Test case with valuesCount = 0
+        val values = ImmutableList.of(1, 2, 3);
+        dataVector = extension.createIntVector(values);
+        val dataCloudArray = new DataCloudArray(dataVector, 0, 0);
+        val array = (Object[]) dataCloudArray.getArray();
+        collector.assertThat(array).isEmpty();
```
mkaufmann, line 136: SHould dataCloudArray.getArray(0,0) work in this scenario?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/DataCloudArray.java
KaviarasuSakthivadivel: Yes, I think we can just return the data here. I will modify it. Thank you for catching this, I was having this as a test.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/DataCloudArray.java
KaviarasuSakthivadivel: Modified the logic, this is simpler. Thank you, Moritz.

mkaufmann (approved):

PR #118, opened 2025-09-30
chore: Automate 3PP version updates

mkaufmann:
We had quite a backlog of outdated versions. This change adds the `version-catalog-update` Gradle plugin to automate the update process. As we recently had problems with reproducible builds due to a 3PP minor update having broken our build this change also switches all libraries to exact versioning. For the special cases where we can't update due to Java 8 / Spark or just migration concerns the logic was explicitly added to block such version candidates.

Drive-By:
- The `DEVELOPMENT.md` is updated to document this (and the existing module graph) commands that should be regularly run to have a healthy code base.
- Explicitly specify Netty version to avoid insecure version pulled in through transitive dependencies
- The updated version of Guava removed the `findbugs:jsr305` transitive dependency and thus the `@Immutable` doc annotation got broken in `EncodingUtils`. As we don't use the annotation anywhere else, we removed it


buildSrc/src/main/kotlin/version-updates.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+
+        // Special cases for libraries where we have to do soft pinning. We don't want to fully pin to still get the latest version for the allowed range.
+        if (("org.junit" in it.candidate.group) && candidateVersion.startsWith("6.")) {
+            false
+        } else if ("org.apache.spark" in it.candidate.group && candidateVersion.startsWith("4.")) {
+            false
+        } else if ("com.squareup.okhttp3" in it.candidate.group && candidateVersion.startsWith("5.")) {
+            false
+        } else if ("org.mockito" in it.candidate.group && candidateVersion.startsWith("5.")) {
+            false
```
vogelsgesang, line 26: More future proof, e.g. as soon as Spark 5 gets published...

```suggestion
        if (("org.junit" in it.candidate.group)) {
           candidateVersion.startsWith("5.")
        } else if ("org.apache.spark" in it.candidate.group) {
            candidateVersion.startsWith("3.")
        } else if ("com.squareup.okhttp3") {
            candidateVersion.startsWith("4.")
        } else if ("org.mockito" in it.candidate.group) {
            candidateVersion.startsWith("3.")
```

buildSrc/src/main/kotlin/version-updates.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+        // For Guava we want to also allow picking the -jre version
+        val candidateVersion = if (it.candidate.version.contains("-jre")) {
+            it.candidate.version.split("-jre")[0]
+        } else {
+            it.candidate.version
+        }
+        val stableKeyword = listOf("RELEASE", "FINAL", "GA").any { candidateVersion.uppercase().contains(it) }
+        val regex = "^[0-9,.v-]+(-r)?$".toRegex()
+
+        // Special cases for libraries where we have to do soft pinning. We don't want to fully pin to still get the latest version for the allowed range.
```
vogelsgesang, line 18: please add a comment to each individual pinned version on why it was pinned

DEVELOPMENT.md
```diff
@@ -4,6 +4,12 @@ Run `./gradlew clean build testCodeCoverageReport` to build the coverage report
 
 There will also be an accompanying CSV in `build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.csv` and a xml file in `build/reports/jacoco/testCodeCoverageReport/testCodeCoverageReport.xml` to be consumed by other tools.
 
+## Maintenance
+
+The following commands should be run regularly to ensure the project stays healthy
+- `./gradlew createModuleGraph`, it updates the module graph in this document and should be run every time the inter project dependencies change
+- `./gradlew versionCatalogUpdate`, should be run every time a dependency is added and regularly anyway to avoid vulnerable versions
```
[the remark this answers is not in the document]
mkaufmann: Thanks for catching that. I checked and it looks like we are completely missing the infrastructure in Gradle. I added a tech debt note to DEVELOPMENT.md to remove the use of those generated files as they seem to provide low value anyway.

PR #127, opened 2025-10-21
chore: Protocol Layer Overhaul for Robustness

mkaufmann:
This PR started with trying to remove the HyperGrpcClientExecutor middle layer. When debugging code it was one of those "what happens in that layer" places that was hard to reason about. Instead there is now a low level `protocol` package that has appropriately named classes that directly expose the required info / results / grpc functionality.. That layer doesn't rely on properties and neither wraps exceptions into SQL Exceptions. This make it generally usable as a low level stand alone API - for example also to fetch JSON results.It provides all the required crucial facilities:
- Proper channel settings to support Hyper message sizes
- Query Info iterator to monitor progress of a query
- Chunk range & row range fetching iterators
- Query Execution & Result consumption iterator (it reuses the chunk range & query info iterators and has less RPC calls and lower latency than the old one)
- Schema access
- ArrowStreamReader conversion

User impacting improvements from this change:
- Previously the whole exception generation logic with customer details had to be wired pretty deeply. With the low level API - that doesn't do the exception conversion - the customer details / query id / sql text context doesn't have to be wired to that level. This concern is now centrally handled top level directly in thel DataCloudConnection / DataCloudStatement class methods. This makes the signatures much easier to read and also it makes it easier to reason around where exceptions have to be converted. For the end user it means that there are now more exceptions where the query id or SQL text are available.
- The new query result iterator code also reduces the number of RPC calls that's required and thus the latency. It keeps the streaming GetQueryInfo call open while fetching results. That way after a chunk is read it is very likely that the next status update is already available and no blocking on network calls is required before fetching the next chunk.
- DataCloudConnection::getSchemaForQueryId would have thrown StatusRuntimeException without catching it
- ByteStringReadableByteChannel was pulling too much data (i.e. resulting in result data getting pulled for schema parsing). With the updated code the initial result set creation and getting access to the query Id + schema can now have significant reduced latency
- The ChunkBased fetching was not logging the timing before, now these timing can be observed from the log
- Adjusting log severity levels. In many cases we used errors even though it was expected / retryable / user caused things. Using info now to make the log more actionable.

Alongside the following more internal smells were addressed:
- Adjust `ElapsedLogger` to use monotonic nanoTime clock approach and extend to allow usage with non throwing measurement blocks.
- Update dataspace header name to new recommendation
- Fix channel leak in `DataCloudConnectionHeaders` test
- Remove leftover stream utilities
- Renamed `ChunkBased` to `ChunkRangeIterator` and `RowBased` to `RowRangeIterator` to make their functionality more clear
- `ArrowStreamReaderCursor` was double wrapping exceptions

Drive-By:
- Enable disabled & broken  AdaptiveResultIteratorFunctional test


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
```diff
@@ -84,7 +80,12 @@ public boolean next() {
             }
             return next;
         } catch (Exception e) {
-            throw new SQLException("Failed to load next batch: " + e.getMessage(), e);
+            // This can happen due ti SneakyThrows.
```
carsheng: ti :D

PR #136, opened 2025-11-13
Avatical Removal Part1: remove Avatica dependency from metadata conversion flow

carsheng:
This change accomplishes Part 1 of removing Avatica from the datacloud-jdbc driver - specifically focusing on marshalling of column metadata. It removes Avatica dependencies from metadata conversion and introduces a custom SimpleResultSet implementation for database metadata queries which includes much of the JDBc boilerplate that was originally handled by Avatica.
QueryMetadataUtil now uses SimpleMetadataResultSet.of() instead of Avatica-based constructors, and type mappings were updated to use standard JDBCType instead of Avatica's SqlType enum.

### DIFF
--- a/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/QueryMetadataUtil.java
+++ b/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/QueryMetadataUtil.java
@@ -197,6 +197,9 @@ final class QueryMetadataUtil {
 
             String typeName = resultSet.getString("datatype");
             typeName = typeName == null ? StringUtils.EMPTY : typeName;
+            // Per PostgreSQL docs (https://www.postgresql.org/docs/current/datatype-numeric.html),
+            // numeric types from Hyper system tables may appear as "numeric", "numeric(p,s)", or
+            // "numeric(p)". Match any variant by checking if the type name contains "numeric".
             if (typeName.toLowerCase().contains("numeric")) {
                 rowData[TYPE_NAME_INDEX] = JDBCType.NUMERIC.toString();
                 rowData[DATA_TYPE_INDEX] = JDBCType.NUMERIC.getVendorTypeNumber();


jdbc-util/src/main/java/com/salesforce/datacloud/jdbc/metadata/SqlType.java
```diff
@@ -0,0 +1,382 @@
...
+    @Getter
+    private final Type type;
+    /// For Numerics: the NUMERIC(precision, scale)
+    /// For Char / Varchar: the length, or 0 for unlimited length
+    /// Unused for other types
+    private final int precisionOrStringLength;
+    /// For Numerics: the NUMERIC(precision, scale)
+    /// Unused for other types
+    private final int scale;
+    /// For Array: the element type
```
mkaufmann: This has very large overlap with com.salesforce.datacloud.jdbc.util.ColumnType. Can you try to use ColumnType instead and extend that where necessary

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/QueryMetadataUtil.java
```diff
@@ -62,22 +55,22 @@ final class QueryMetadataUtil {
...
-            immutableEntry("time", SqlType.TIME.toString()),
-            immutableEntry("timetz", SqlType.TIME.toString()),
-            immutableEntry("timestamp", SqlType.TIMESTAMP.toString()),
-            immutableEntry("timestamptz", SqlType.TIMESTAMP.toString()),
-            immutableEntry("array", SqlType.ARRAY.toString()));
+            immutableEntry("int2", JDBCType.SMALLINT.toString()),
+            immutableEntry("int4", JDBCType.INTEGER.toString()),
+            immutableEntry("oid", JDBCType.BIGINT.toString()),
+            immutableEntry("int8", JDBCType.BIGINT.toString()),
+            immutableEntry("float", JDBCType.DOUBLE.toString()),
```
carsheng: @mkaufmann confirm whether FLOAT --> double conversion is correct?

mkaufmann: It's pretty unclear where the names here are coming from as they don't seem to be SQL types. The only use case of this map seems to be for `    private static List<Object> constructColumnData(ResultSet resultSet) throws SQLException {`

That method doesn't really make sense though as it seems to be only used for convering a ResultSet in the end to a result set. So I think this + constructColumnData can just be completely removed. Instead of all that complication instead this function
```
    public static ResultSet createColumnResultSet(
            String schemaPattern, String tableNamePattern, String columnNamePattern, Connection connection)
            throws SQLException {

        final List<Object> data;

        try (val statement = connection.createStatement()) {
            val getColumnsQuery = getColumnsQueryInner(schemaPattern, tableNamePattern, columnNamePattern);
            val resultSet = statement.executeQuery(getColumnsQuery);
            data = constructColumnData(resultSet);
        }

        return getMetadataResultSet(QueryDBMetadata.GET_COLUMNS, data);
    }
```
Should be adjusted so that it directly return the resultSet from `statement.executeQuery(getColumnsQuery);` and to enable that the query needs to be adjusted to directly return the columns with the expected names and types for `GET_COLUMNS`. That can be achieved by using `CAST( ... AS <TYPE>)` and ` as <COLUMN_NAME>` in the top level select

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/ColumnType.java
```diff
@@ -40,20 +43,31 @@ public ColumnType(JDBCType type) {
         this.arrayElementType = null;
         this.precisionOrStringLength = -1;
         this.scale = 1;
+        nullable = true;
```
mkaufmann: Does it make sense to have a default? Shouldn't this be passed in explicitly?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/ColumnType.java
```diff
@@ -251,4 +265,39 @@ public int getDisplaySize() {
                 return getPrecisionOrStringLength();
         }
     }
+
+    // See table B.3 from https://download.oracle.com/otn-pub/jcp/jdbc-4_3-mrel3-eval-spec/jdbc4.3-fr-spec.pdf
+    public String getJavaTypeName() {
```
mkaufmann: Would adjust the interface to just return the Java class here (and rename method to `getJavaType()` and then in getColumnClassName call getJavaType().getName()

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/SimpleResultSetMetaData.java
```diff
@@ -0,0 +1,178 @@
...
+
+    @Override
+    public String getTableName(int column) throws SQLException {
+        return "";
+    }
+
+    @Override
+    public String getCatalogName(int column) throws SQLException {
+        return "";
+    }
```
mkaufmann: Isn't this a regression compared to today? I would expect that these fields are important for DatabaseMetaData.getColumns use cases

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/QueryMetadataUtil.java
```diff
@@ -62,22 +55,22 @@ final class QueryMetadataUtil {
...
+            immutableEntry("float4", JDBCType.REAL.toString()),
+            immutableEntry("float8", JDBCType.DOUBLE.toString()),
+            immutableEntry("bool", JDBCType.BOOLEAN.toString()),
+            immutableEntry("char", JDBCType.CHAR.toString()),
+            immutableEntry("text", JDBCType.VARCHAR.toString()),
+            immutableEntry("date", JDBCType.DATE.toString()),
+            immutableEntry("time", JDBCType.TIME.toString()),
```
mkaufmann: Should this be part of ColumnMetadata given that it has a typename and isn't this the mapping from type name to JDBC Type?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/QueryMetadataUtil.java
```diff
@@ -96,14 +89,8 @@ public static ResultSet createTableResultSet(
         return getMetadataResultSet(QueryDBMetadata.GET_TABLES, data);
     }
 
-    static AvaticaResultSet getMetadataResultSet(QueryDBMetadata queryDbMetadata, List<Object> data)
-            throws SQLException {
-        QueryResultSetMetadata queryResultSetMetadata = new QueryResultSetMetadata(queryDbMetadata);
-        List<ColumnMetaData> columnMetaData = convertJDBCMetadataToAvaticaColumns(queryResultSetMetadata);
```
mkaufmann: Can this method be deleted `convertJDBCMetadataToAvaticaColumns`?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/SimpleResultSetMetaData.java
carsheng: This actually mimics the current behavior on the driver.  (Altho the old methods return String.EMPTY), this wouldn't be a regression. 
These methods (getCatalogName(), getSchemaName() ...etc) for a specific column are considered lineage methods which we don't currently support - i.e. a column does not know what schema or catalog it resides in.  ResultSetMetadata only contains info about types and properties of the columns, not anything beyond them

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/SimpleResultSetMetaData.java
```diff
@@ -0,0 +1,178 @@
...
+ * This file is part of https://github.com/forcedotcom/datacloud-jdbc which is released under the
+ * Apache 2.0 license. See https://github.com/forcedotcom/datacloud-jdbc/blob/main/LICENSE.txt
+ */
+package com.salesforce.datacloud.jdbc.core.metadata;
+
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import lombok.val;
+
+public class SimpleResultSetMetaData implements ResultSetMetaData {
```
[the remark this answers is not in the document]
carsheng: Good catch thanks Moritz. updated

PR #140, opened 2025-12-02
feat: Improve authentication error logging and retry handling

KaviarasuSakthivadivel:
Enhanced error logging and retry logic for authentication failures to provide better visibility into transient server errors and improve resilience against flaky authentication endpoints.

Changes:
- Enhanced AuthorizationException.getMessage() to include HTTP status code and error description in a structured format
- Added AuthorizationException.isRetriable() method to identify transient errors that should be retried:
  * All 5xx server errors (500, 502, 503, 504, etc.)
  * 429 rate limit errors
- Modified retry policy in DataCloudTokenProvider to retry retriable AuthorizationExceptions while failing fast on permanent auth errors

This improves debugging of authentication issues and makes the driver more resilient to transient server-side failures commonly seen in test environments.

### DIFF
--- a/jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/errors/AuthorizationException.java
+++ b/jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/errors/AuthorizationException.java
@@ -40,9 +40,9 @@ public class AuthorizationException extends Exception {
 
     /**
      * Determines if this exception represents a transient error that should be retried.
-     * Retries are appropriate for:
-     * - 5xx server errors (500, 502, 503, 504)
-     * - 400 errors with retry hints in the error description (e.g., "retry your request", "unknown_error")
+     *
+     * Retriable errors: 5xx server errors and 429 Too Many Requests.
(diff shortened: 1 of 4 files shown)


jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/errors/AuthorizationException.java
```diff
@@ -13,4 +13,71 @@ public class AuthorizationException extends Exception {
...
+            // Retry on 400 errors that indicate transient issues
+            if (httpCode == 400) {
+                String description = errorDescription != null ? errorDescription.toLowerCase() : "";
+                // Check for retry hints in the error description
+                return description.contains("retry")
+                        || description.contains("unknown_error")
+                        || description.contains("temporary")
+                        || description.contains("rate limit")
+                        || description.contains("throttle");
+            }
```
mkaufmann: Did you see this really occurring? Generally 400 looks to be something that the server consider a client error and thus I wouldn't retry on it https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/400.

carsheng (approved):

PR #142, opened 2025-12-10
fix: Breaking - Remove data loss for slow readers

mkaufmann:
The driver had a bug where an extremely slow reading client (>120 seconds to consume all chunks from ExecuteQuery) would see data loss. All the remaining unconsumed messages from the first chunk would silently get missed. The root cause is a combination of graceful handling of the adaptive flow which cancels the call and that gRPC framework let's errors race messages (see JDBCLimitsTest.testSlowReader for a test case the reproduced the data loss and is fixed now) This change is working around this problem by integrating on a lower level of the gRPC framework and thus being able to control so that these races won't happen.

Breaking Part `HyperGrpcStubProvider`: To switch to the lower level gRPC APIs the driver is now using Async Stubs internall and not BlockingStubs. Thus consumers that create their own Channels & Stubs through the StubProvider had to adjust their code to now produce an Async stub.

Technical Details:
The new classes in `com.salesforce.datacloud.jdbc.protocol.grpc.util` implement the low level streaming call flow. As the JDBC driver is inherently synchronous we have to create a bridge from the async StreamObserver interface to an iterator. This bridge is implemented in the `BufferingStreamObserver` and the `BufferingStreamIterator` classes. To minimize general HTTP Stream / Connection head of line blocking issues due to slow readers causing messages not being consumed in timely fashion we buffer the messages from the streaming gRPC call within the driver. This way we guarantee that data is read by the driver at network speed. Memory wise the total streaming response size is limited by the protocol so from a memory perspective this will not result in OOM risk. By using the lower level interface we can now properly close the calls in case we finish early. The `waitFor` now cancels the streaming call if the predicate is met.

Drive-By
- Move all RPC time tracking directly within the observer to ensure consistent time capturing for streaming calls. To facilitate that the start / fail / success messages of the `ElapsedLogger` were exposed and the now unused nonThrowing variant removed. It now also includes response size tracking and the "millis" reporting was fixed as well
- Remove unused functionality in jdbc-util
- Add support to override request-timeout in the HyperProcessConfig for testing, fixed the logic for generating the arguments to properly support multiple ones
- Clean up the HyperGrpcExecutorClient. It's only real responsibility (after removing remaining stub usages) is to create QueryParams. Thus it is now renamed to ExecuteQueryParamBuilder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,1 @@
 # Changelog
-
-## [0.42.0] - 2025-11-11
-
-### Added
-- Initial automated release pipeline setup
-
-[0.42.0]: https://github.com/forcedotcom/datacloud-jdbc/releases/tag/0.42.0
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,4 +8,4 @@ hyperApiVersion=0.0.23576.r0633e4a4
 
 # Revision here is what is used when producing local and snapshot builds,
 # it is ignored for releases which instead use the tag
-revision=0.42.0
+revision=0.41.0  # x-release-please-version
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -1,11 +1,11 @@
 {
   "packages": {
     ".": {
-      "release-type": "java",
-      "version-file": "gradle.properties",
-      "version-file-prefix": "revision=",
+      "release-type": "simple",
       "changelog-path": "CHANGELOG.md",
-      "extra-files": []
+      "extra-files": [
+        "gradle.properties"
+      ]
     }
   },
   "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json"


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/grpc/util/ClosableCancellingStreamObserver.java
```diff
@@ -0,0 +1,71 @@
...
+     * Receives the {@link ClientCallStreamObserver} before the call starts. If a prior
+     * {@link #close()} was requested, the call is cancelled immediately to avoid starting
+     * a new stream after shutdown.
+     *
+     * @param callStream the call stream observer provided by gRPC
+     */
+    @Override
+    public void beforeStart(ClientCallStreamObserver<ReqT> callStream) {
+        this.callStream = callStream;
+        // Request many messages to avoid backpressure, with 16 outstanding requests we should have enough buffering
```
KaviarasuSakthivadivel: I'm curious to understand why 16 here?

afurdell (approved):

praveen2450 (approved):

PR #166, opened 2026-04-06
feat: Avatical Removal Part 3/3 - Remove Avatica dependency completely

carsheng:
Completes the removal of Apache Calcite Avatica from jdbc-core. The driver no longer depends on Avatica for result set handling, cursor management, type accessors                                  
                                                                                                                                                                      
  - Remove Avatica from StreamingResultSet: no longer extends AvaticaResultSet. Implements DataCloudResultSet directly with all JDBC ResultSet methods delegating to QueryJDBCAccessor instances. Adds proper checkClosed() guards on all methods.
  - Remove Avatica cursor layer: delete MetadataCursor (extended AbstractCursor) and MetadataResultSet (extended AvaticaResultSet). Replace with DataCloudMetadataResultSet, a standalone metadata result set implementation.                                                                                                                                    
  - Extract DataCloudQueryAccessor interface: defines the accessor contract (all typed getters) that QueryJDBCAccessor implements, replacing the dependency on Avatica's AbstractCursor.Accessor.                                                                                                                                  
  - Rename for clarity: SimpleResultSetMetaData → DataCloudResultSetMetaData, SimpleMetadataResultSet → DataCloudMetadataResultSet, DataCloudAccessor →  DataCloudQueryAccessor.


jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/metadata/SimpleResultSetMetaDataTest.java
```diff
@@ -137,7 +137,7 @@ public void getColumnTypeName() throws SQLException {
 
     @Test
     public void testIsReadOnly() throws SQLException {
-        assertThat(simpleResultSetMetaData.isReadOnly(1)).isTrue();
+        assertThat(simpleResultSetMetaData.isReadOnly(1)).isFalse();
```
mkaufmann: This looks wrong

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/QueryJDBCDataCursorTest.java
```diff
@@ -1,34 +0,0 @@
...
-    }
-
-    @Test
-    public void testCursorNextAndRowCount() {
-        List<Object> data = new ArrayList<>();
-        data.add("Account_home_dll");
-        cursor = new MetadataCursor(data);
-        cursor.next();
-        assertThat(cursor.next()).isFalse();
-    }
```
mkaufmann, line 33: Do we have test coverage for cursor somewhere else?

mkaufmann (approved):

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/DataCloudAccessor.java
```diff
@@ -0,0 +1,94 @@
...
+import java.sql.SQLException;
+import java.sql.SQLXML;
+import java.sql.Struct;
+import java.sql.Time;
+import java.sql.Timestamp;
+import java.util.Calendar;
+import java.util.Map;
+
+/**
+ * Project-owned replacement for Avatica's {@code Cursor.Accessor} interface. Provides the same
```
mkaufmann: Remove reference to avatica

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/DataCloudResultSetMetaData.java
```diff
@@ -58,7 +59,12 @@ public String getColumnLabel(int column) throws SQLException {
 
     @Override
     public int getColumnType(int column) throws SQLException {
-        return getColumn(column).getType().getType().getVendorTypeNumber();
+        int typeId = getColumn(column).getType().getType().getVendorTypeNumber();
+        // Arrow timestamps with timezone map to JDBC TIMESTAMP, not TIMESTAMP_WITH_TIMEZONE
+        if (typeId == Types.TIMESTAMP_WITH_TIMEZONE) {
+            return Types.TIMESTAMP;
+        }
```
mkaufmann: Why do we need to adjust that here? Feels wrong to me to override the column type

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
```diff
@@ -80,7 +79,6 @@ public boolean next() {
             }
             return next;
         } catch (Exception e) {
```
mkaufmann, line 76: Why remove comment

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ColumnNameResolver.java
```diff
@@ -35,25 +35,17 @@ public final class ColumnNameResolver {
...
-
-        // Second pass: index lowercase labels to ordinals, but only if the lowercase key doesn't exist yet
-        // This ensures the first column with a given lowercase name wins (lowest ordinal)
-        for (ColumnMetaData columnMetaData : columns) {
-            if (columnMetaData.label != null) {
-                String lowerLabel = columnMetaData.label.toLowerCase();
-                // Only add if this lowercase key hasn't been seen yet (preserves first occurrence)
-                lowercaseColumnLabelMap.putIfAbsent(lowerLabel, columnMetaData.ordinal);
+                exactColumnLabelMap.putIfAbsent(label, i);
+                lowercaseColumnLabelMap.putIfAbsent(label.toLowerCase(), i);
```
mkaufmann, line 50: Doesn#t this change behavior? I think this should break a test case

KaviarasuSakthivadivel: Do we need to change this behaviour?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/DataCloudAccessor.java
```diff
@@ -0,0 +1,94 @@
...
+ * method signatures for column value access without any Avatica dependency.
+ */
+public interface DataCloudAccessor {
```
KaviarasuSakthivadivel: DataCloudQueryAccessor?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/QueryJDBCAccessor.java
```diff
@@ -23,9 +23,8 @@
 import java.util.Calendar;
 import java.util.Map;
 import java.util.function.IntSupplier;
-import org.apache.calcite.avatica.util.Cursor.Accessor;
 
-public abstract class QueryJDBCAccessor implements Accessor {
+public abstract class QueryJDBCAccessor implements DataCloudAccessor {
     private final IntSupplier currentRowSupplier;
     protected boolean wasNull;
     protected QueryJDBCAccessorFactory.WasNullConsumer wasNullConsumer;
```
mkaufmann: Is this wasNullConsumer even still required without avatica? I think this can be removed

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -4,144 +4,473 @@
...
             String queryId,
-            AvaticaStatement statement,
-            QueryState state,
-            Meta.Signature signature,
-            ResultSetMetaData resultSetMetaData,
-            TimeZone timeZone,
-            Meta.Frame firstFrame)
-            throws SQLException {
-        super(statement, state, signature, resultSetMetaData, timeZone, firstFrame);
+            SimpleResultSetMetaData metadata,
```
mkaufmann: Can we just rename this to DataCloudResultSetMetaData?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
```diff
@@ -35,27 +32,30 @@ class ArrowStreamReaderCursor extends AbstractCursor {
...
+
     private final AtomicInteger currentIndex = new AtomicInteger(INIT_ROW_NUMBER);
 
     private void wasNullConsumer(boolean wasNull) {
-        this.wasNull[0] = wasNull;
+        this.wasNull = wasNull;
+    }
+
+    boolean wasNull() {
+        return wasNull;
```
mkaufmann: Who is using this?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/SimpleResultSetMetaData.java
```diff
@@ -142,8 +148,9 @@ public String getCatalogName(int column) throws SQLException {
 
     @Override
     public boolean isReadOnly(int column) throws SQLException {
-        // Our result sets are always read only
-        return true;
+        // Match standard JDBC behavior: readOnly=false does not mean the column is writable,
+        // it means we don't know. isWritable() and isDefinitelyWritable() still return false.
+        return false;
```
mkaufmann: I think the old behavior was correct because we know that this column can't be written through SQL as DC only supports SELECT clauses

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ColumnNameResolver.java
mkaufmann: Actually taking a second look I think the new logic is correct. On the lookup side we always first check exact and then lower case. Having a single pass is more efficient than interating twice.

Please keep the documentation on the semantics of the two maps and how this is working.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -4,144 +4,473 @@
...
-    public int getRow() {
+    public int getRow() throws SQLException {
+        checkClosed();
         return cursor.getRowsSeen();
     }
 
+    @Override
+    public ResultSetMetaData getMetaData() throws SQLException {
+        return metadata;
+    }
```
mkaufmann, line 147: We should also disallow this when the resultset ist closed 
```suggestion
    @Override
    public ResultSetMetaData getMetaData() throws SQLException {
        checkClosed();
        return metadata;
    }
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -4,144 +4,473 @@
...
+
     @Override
     public int findColumn(String columnLabel) throws SQLException {
         return columnNameResolver.findColumn(columnLabel);
     }
```
mkaufmann, line 153: We should also disallow this when the resultset is closed

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -4,144 +4,473 @@
...
-     * Override getter methods that take String columnLabel to ensure they use our optimized findColumn() method.
-     * This is necessary because Avatica's implementation might use a private method or cache.
-     */
     @Override
-    public String getString(String columnLabel) throws SQLException {
-        int columnIndex = findColumn(columnLabel);
-        return getString(columnIndex);
+    public Statement getStatement() throws SQLException {
+        return null;
+    }
```
mkaufmann, line 159: We should also allow this when the result set is closed, here and further methods

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/DataCloudAccessor.java
carsheng: fixed

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
carsheng: fixed

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/QueryJDBCAccessor.java
carsheng: Removed wasNullConsumer pattern across accessors

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/metadata/SimpleResultSetMetaDataTest.java
carsheng: See https://github.com/forcedotcom/datacloud-jdbc/pull/166#discussion_r3076221817 other thread for continued discussion

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/SimpleResultSetMetaData.java
carsheng: At first i thought so too, but i've since realized the following the "old behavior" was actually incorrect.

According to QueryJDBCReferenceTest, which compares behavior against the Postgres driver, isReadOnly=false for a **column** is in line w/ what Postgres driver does. 
So the correct behavior should be:
```
Generally:
ResultSet.isReadonly=true // the resultset itself isReadonly, we don't support write operations
For this line:
ResultSet.getMetadata.isReadOnly(1)=false //a column is technically not readonly
```

**Why was this not caught earlier?** 
SimpleResultSetMetadata in the earlier avatica PR (aka, "old behavior") were limited to only the metadata flow. Now that that SimpleResultSetMetadata is now incorporated into the normal ResultSet flow  (i.e. u make a query, then look at the metadata), QueryJDBCReferenceTest fails b/c it expects column.isReadonly=false. See here in reference.json: https://github.com/forcedotcom/datacloud-jdbc/blob/main/jdbc-reference/src/main/resources/reference.json#L19

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
carsheng: Removed as wasNullConsumer pattern is no longer relevant

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/DataCloudResultSetMetaData.java
carsheng: It does feel wrong, however the reason for this right now is that Arrow's TIMESTAMPSECTZ / TIMESTAMPMILLITZ / etc. store an epoch value + a timezone string, but the TimeStampVectorAccessor already normalizes these to UTC LocalDateTime / Timestamp values before returning them. By the time the JDBC caller gets the value via getTimestamp(), the timezone has already been folded into the instant — the returned java.sql.Timestamp doesn't carry timezone info.

If we remove the override, and TimeStampVectorAccessor remains as is, JDBCReferenceTest fails because it expects `select timestamptz '1000-01-01 01:00:00+0100'` to map to `columnType=93` and `className java.sql.TimeStamp`

My view is that this is a symptom of the fact we don't currently correctly handle timestamp/timezone types in the JDBC driver, and that this should go away once #158 gets in

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
carsheng: done

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
carsheng: done

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/DataCloudAccessor.java
carsheng: done

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/SimpleResultSetMetaData.java
mkaufmann: >isReadOnly=false for a column is in line w/ what Postgres driver does.

Postgres driver is setting this to falser because one can do insert / update statements in Postgres. In Data Cloud that doesn't work so for us it would be correct to set isReadOnly true AFAIK

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/DataCloudResultSetMetaData.java
mkaufmann: @KaviarasuSakthivadivel / @carsheng let's discuss on how we want to sequence the two PRs.

KaviarasuSakthivadivel (approved):

mkaufmann (approved):

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/SimpleResultSetMetaData.java
carsheng: Fix made so that readOnly=true in all cases.

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/QueryJDBCDataCursorTest.java
carsheng: Yes. This test was deleted as it was minimal and also was dependent on MetadataCursor which extended Avatica's AbstractCursor. DataCloudMetadataResultSetTest.java methods testGetRow() and testNext() cover the cases tested here.

KaviarasuSakthivadivel (approved):

PR #170, opened 2026-04-21
fix: async interrupt race condition in SyncIteratorAdapter

mkaufmann:
Fixes IllegalStateException("Unfulfilled previous future when next is requested") thrown from AsyncStreamObserver.requestNext() when a thread is interrupted while blocked on a gRPC response.

Root cause: the old interrupt handler recursively called hasNext(), which requested a *new* future from the async iterator before gRPC's async onError callback had resolved the *previous* pendingFuture.

Fix: hoist the CompletableFuture out of the while loop so that on interrupt we close() the iterator (triggering gRPC cancellation) and then re-wait on the *same* future. The real gRPC CANCELLED error then propagates naturally through the ExecutionException path.

Adds a regression test that simulates real gRPC behavior (close() does not synchronously complete the future) and asserts the cancellation error surfaces instead of the IllegalStateException.

### DIFF
--- a/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/async/core/SyncIteratorAdapter.java
+++ b/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/async/core/SyncIteratorAdapter.java
@@ -8,6 +8,7 @@ import com.salesforce.datacloud.jdbc.protocol.CloseableIterator;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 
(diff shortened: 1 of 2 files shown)


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/async/core/SyncIteratorAdapter.java
```diff
@@ -61,25 +62,26 @@ public boolean hasNext() {
...
                     interrupted = true;
-                    // This will cause the ongoing call to be stopped and thus the future will get an error element
-                    // which will cause the while loop to exit.
                     try {
                         asyncIterator.close();
                     } catch (Exception ignore) {
                     }
-                    return hasNext();
                 } catch (ExecutionException ee) {
                     Throwable cause = ee.getCause();
```
KaviarasuSakthivadivel, line 93: I think we need to add `done = true;` here right? When a thread is interrupted and gRPC cancellation propagates, hasNext() throws a RuntimeException (the CANCELLED error) to the caller. At that point the gRPC stream is permanently dead but done is still false.

KaviarasuSakthivadivel: Also, I think we can collapse this catch block. 

```
 catch (ExecutionException | CompletionException ee) {
      Throwable cause = ee.getCause();
      if (cause instanceof RuntimeException) {
          throw (RuntimeException) cause;
      }
      throw new RuntimeException(cause);
  }

```

mkaufmann: Great catch!

KaviarasuSakthivadivel (approved):

carsheng (approved):

PR #173, opened 2026-04-23
test: retry ShadedJarIntegrationTest on transient OAuth 400

mkaufmann:
## Summary
- `ShadedJarIntegrationTest.testConnectionAndQueryExecution` has been the single remaining flaky test on `main` across the last 50 commits (4 known failures in the sampled window: runs 24843852285, 22970082839, 22968279868, 21726987152).
- Root cause: `login.test2.pc-rnd.salesforce.com` intermittently returns `HTTP 400 {"error":"unknown_error","error_description":"retry your request"}`. The driver's `AuthorizationException.isRetriable()` only retries `5xx` and `429`, so this transient surfaces as a hard failure.
- Fix: wrap `DriverManager.getConnection` in a bounded exponential-backoff loop (up to ~60s total) that retries **only** when the exception chain contains both `HTTP 400` and `retry your request`. Every other `SQLException` still propagates immediately, so real regressions are not masked.
- This is test-environment instability: per HTTP semantics a 400 is non-retriable, so the driver is correct not to
retry it. Scope the workaround to this integration test with a bounded exponential-backoff loop (up to ~60s) that retries only when the exception chain contains both "HTTP 400" and "retry your request". Every other SQLException still propagates immediately. If this same response is ever observed outside the test environment, the upstream service should be changed to return an appropriate retriable status (5xx / 429) and the driver's core retry policy will handle it
## Test plan
- [x] `./gradlew spotlessApply :integration-test:compileTestJava` passes locally
- [x] CI `Build, test, and publish snapshot` succeeds on this PR
- [ ] Re-trigger the workflow a few times to confirm the retry absorbs the transient without masking other failures

### DIFF
--- a/integration-test/src/test/java/com/salesforce/datacloud/jdbc/integration/ShadedJarIntegrationTest.java
+++ b/integration-test/src/test/java/com/salesforce/datacloud/jdbc/integration/ShadedJarIntegrationTest.java
@@ -101,7 +101,13 @@ public class ShadedJarIntegrationTest {
         log.info("  Attempting connection to: {}", jdbcUrl);
 
         // Test connection creation and query execution - this will fail if gRPC service files are broken
-        try (Connection conn = DriverManager.getConnection(jdbcUrl, props)) {
+        // The test OAuth endpoint at login.test2.pc-rnd.salesforce.com intermittently returns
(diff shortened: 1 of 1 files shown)


KaviarasuSakthivadivel (approved):

j10t (approved):
LGTM - and thanks for the solid PR description!

PR #174, opened 2026-04-24
refactor: centralize type mapping to fix metadata-path divergence

mkaufmann:
## Why

Column type information currently flows through three independent paths — Arrow schema, `pg_catalog.format_type()`, and Spark — each with its own per-site mapping logic. That scatter has caused repeated, silently-diverging bugs: `pg_catalog` rows returning raw PostgreSQL OIDs because `QueryMetadataUtil.dbTypeToSql` keyed on `int2`/`int4`/`int8` aliases that Hyper never emits; `ResultSetMetaData` disagreeing with `DatabaseMetaData.getColumns()` on type names and codes for the same column; DECIMAL vs NUMERIC inconsistency between the two paths; `getTypeInfo()` returning an empty result; engine-level concerns (parameter binding, vector population) reaching into JDBC exception types.

## Observable behavior changes

**New capabilities / bug fixes:**
- `DatabaseMetaData.getTypeInfo()` returns a real 18-column result enumerated from `HyperTypeKind` (was empty before).
- Prepared-statement parameter Arrow round-trip: CHAR(1), CHAR(n), and VARCHAR(n) now round-trip through Arrow losslessly via `hyper:type` / `hyper:max_string_length` field metadata, matching the convention Hyper's server already uses on the inbound side.
- `DatabaseMetaData.getColumns().COLUMN_SIZE` and `DECIMAL_DIGITS` are now derived from the parsed type's declared precision/scale (e.g. `numeric(10,5)` reports `COLUMN_SIZE=10, DECIMAL_DIGITS=5`; `char(1)` reports `COLUMN_SIZE=1`) instead of being hardcoded to `255` / `2`.
- `DatabaseMetaData.getColumns().DATA_TYPE` is always a `java.sql.Types` value (was a raw PG OID for unmapped types); unsupported types surface as `Types.OTHER` with the raw name preserved via `HyperTypeKind.UNKNOWN` rather than throwing.
- `DECIMAL_DIGITS` stays `0` for `FLOAT4` / `FLOAT8` (binary floats have no meaningful decimal scale) instead of being hardcoded to `2`.

**Syntactic / naming changes:**
- `DatabaseMetaData.getColumns().TYPE_NAME` is consistently JDBC-uppercase (`"BIGINT"`, `"INTEGER"`, `"DECIMAL"`, …) instead of a mix of Hyper-lowercase and uppercase. `INTERVAL` and `JSON` show the Hyper-uppercase form; unmodeled system-catalog types (e.g. `aclitem`) show the raw Hyper name.
- Numeric/decimal unified to `DECIMAL` (both Arrow Decimal and Hyper `numeric` map to `HyperTypeKind.DECIMAL` / `JDBCType.DECIMAL`).

## What's introduced

Introduce `HyperType`, a Hyper-native internal type model, and route every Arrow, `pg_catalog`, prepared-statement, and Spark conversion through it. The data layer lives under a new `com.salesforce.datacloud.jdbc.protocol.data` package that has no JDBC dependency, so engine-level callers can use it without a JDBC import; the JDBC-bound helpers stay in `com.salesforce.datacloud.jdbc.core.types.HyperTypes`. Translation to `java.sql.JDBCType` / `Types` only happens at the JDBC API boundary.

**New in `protocol.data` (JDBC-free):**
- `HyperTypeKind` + `HyperType` value class with a visitor and native `toHyperTypeName()`. `HyperTypeKind.UNKNOWN` preserves the raw Hyper type name for system-catalog types the driver doesn't model (e.g. `aclitem`).
- `PgCatalogTypeParser`, `ArrowToHyperTypeMapper`, `HyperTypeToArrow` (format-boundary converters).
- `ColumnMetadata` with optional `typeName` override.
- `ParameterBinding` + `ParameterAccumulator` (replaces `ParameterManager`); throws `IllegalArgumentException` so it is usable without JDBC.
- `ArrowUtils`, `VectorPopulator` (moved from `.util`).

**`HyperTypes` stays in `core.types`** as the JDBC-bound boundary: `toJdbcType*`, `fromJdbcTypeCode`, `toJavaClass`, `getPrecision`/`getScale`/`getDisplaySize`, `isSigned`/`isCaseSensitive`/`isNullable`, `needsDecimalDigits`, `needsCharOctetLength`, `typeInfoRows`, plus the `isIntegerLike` / `isStringLike` helpers that wrap exhaustive enum switches so callers can do a family check without replicating the switch.

## Structural details

**Deleted:** `ColumnType`, `ArrowToColumnTypeMapper`, the `QueryMetadataUtil.dbTypeToSql` + `isDecimalType`/`isCharType` helpers, the `ArrowUtils.SQL_TYPE_TO_FIELD_TYPE` map.

**Migrated callers:** `ColumnMetadata` holds `HyperType`; `DataCloudResultSetMetaData`, `StreamingResultSet`, `DataCloudArray`, `DataCloudConnection.getSchemaForQueryId`, and `ArrowUtils.toColumnMetaData` go through `ArrowToHyperTypeMapper` + `HyperTypes`; `QueryMetadataUtil.constructColumnData` parses via `PgCatalogTypeParser` and falls back to `HyperType.unknown(raw)` for unmodeled system-catalog types; `SimpleResultSet` switches on `HyperTypeKind` via the new helpers (and picks up TINYINT numeric coercions and the `getFloat` wasNull/range-check fix that were planned as Phase 0 work); `DataCloudPreparedStatement` wraps `IllegalArgumentException` from `ParameterAccumulator` into `SQLException`; `VectorPopulator` is keyed by `HyperTypeKind`; Spark `TypeMapping` handles TINYINT, NUMERIC, CHAR, NULL, FLOAT, NCHAR, NVARCHAR, LONGNVARCHAR.

**Follow-ups noted in code / comments (not in this PR):**
- `SimpleResultSet` is documented for removal once `DataCloudMetadataResultSet` migrates to `StreamingResultSet` — one result-set implementation instead of two.
- `QueryResultArrowStream.asArrowStreamReader`'s `RootAllocator` lifetime needs a dedicated fix (it flows into `ArrowStreamReader`, which doesn't close it).
- `TIME WITH TIME ZONE` Arrow round-trip is pinned as intended asymmetry — offset honored on input but not preserved on output, matching Hyper's read-side behavior. Proper round-trip would need either a synthetic Arrow schema or Hyper-protocol coordination.


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/resultset/SimpleResultSet.java
```diff
@@ -299,9 +302,10 @@ public Object getObject(int columnIndex) throws SQLException {
             case CHAR:
             case VARCHAR:
                 return getString(columnIndex);
+            default:
+                throw new SQLException("Unsupported column type in `getObject`: "
+                        + metadata.getColumn(columnIndex).getType().toString());
```
mkaufmann: Don't handle default so that when a type is added a linter can tell that it should get handling here

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/DefaultParameterAccumulator.java
```diff
@@ -0,0 +1,33 @@
...
+ * Apache 2.0 license. See https://github.com/forcedotcom/datacloud-jdbc/blob/main/LICENSE.txt
+ */
+package com.salesforce.datacloud.jdbc.protocol.data;
+
+import java.util.ArrayList;
+import java.util.List;
+import lombok.Getter;
+
+@Getter
+public class DefaultParameterAccumulator implements ParameterAccumulator {
```
mkaufmann: Why have a separate interface and accumulator. Do we really need that or can we jsut rename DefaultParameterAccumulator to ParameterAccumulator and drop the current interface

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
```diff
@@ -0,0 +1,598 @@
...
+            case BOOL:
+                return JDBCType.BOOLEAN;
+            case INT8:
+                return JDBCType.TINYINT;
+            case INT16:
+                return JDBCType.SMALLINT;
+            case INT32:
+                return JDBCType.INTEGER;
+            case INT64:
+            case OID:
```
mkaufmann, line 58: Do we really need to map OID to bigint? It's more of an internal type that I'm not sure if we want to keep exposed and has special semantics. Would it be enough to map to other?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
```diff
@@ -0,0 +1,598 @@
...
+            case DECIMAL:
+                return JDBCType.DECIMAL;
+            case CHAR:
+                return JDBCType.CHAR;
+            case VARCHAR:
+                return JDBCType.VARCHAR;
+            case BINARY:
+                return JDBCType.BINARY;
+            case VARBINARY:
+                return JDBCType.VARBINARY;
```
mkaufmann, line 78: Are both of these types used by Hyper and available in Arrow? Or is binary rather unused (should we even have it in HyperTeps then?)

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
```diff
@@ -0,0 +1,598 @@
...
+            case Types.TIMESTAMP:
+                return HyperType.timestamp(nullable);
+            case Types.TIMESTAMP_WITH_TIMEZONE:
+                return HyperType.timestampTz(nullable);
+            case Types.ARRAY:
+                return HyperType.array(HyperType.varcharUnlimited(nullable), nullable);
+            case Types.NULL:
+                return HyperType.nullType();
+            default:
+                throw new IllegalArgumentException("Unsupported JDBC type code: " + jdbcTypeCode);
```
mkaufmann: Also here rather prefer to not have default case and have the throw outside. Also adjust other places in the PR

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
```diff
@@ -0,0 +1,598 @@
...
+    public static String toJdbcTypeName(HyperType t) {
+        switch (t.getKind()) {
+            case INTERVAL:
+                return "INTERVAL";
+            case JSON:
+                return "JSON";
+            case UNKNOWN:
+                return t.getUnknownTypeName();
+            default:
+                return toJdbcType(t).getName();
```
mkaufmann, line 182: This looks like a valid case of default

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
```diff
@@ -0,0 +1,598 @@
...
+            case CHAR:
+            case VARCHAR:
+            case BINARY:
+            case VARBINARY:
+            case DATE:
+                return UNKNOWN_SCALE;
+            case FLOAT4:
+            case FLOAT8:
+                // Matches legacy behaviour: floating-point scale equals precision.
+                return getPrecision(t);
```
mkaufmann, line 312: This comment does not help. Please check if this is to match Postgres JDBC behavior and update comment

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
```diff
@@ -0,0 +1,598 @@
...
+            case INT64:
+                return HyperType.int64(true);
+            case OID:
+                return HyperType.oid(true);
+            case FLOAT4:
+                return HyperType.float4(true);
+            case FLOAT8:
+                return HyperType.float8(true);
+            case DECIMAL:
+                return HyperType.decimal(38, 0, true);
```
mkaufmann, line 614: Should the prototype also have non zero digits?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudDatabaseMetadata.java
```diff
@@ -753,7 +755,8 @@ public ResultSet getCrossReference(
 
     @Override
     public ResultSet getTypeInfo() throws SQLException {
-        return DataCloudMetadataResultSet.empty();
+        return DataCloudMetadataResultSet.of(
+                new DataCloudResultSetMetaData(MetadataSchemas.TYPE_INFO), HyperTypes.typeInfoRows());
```
mkaufmann, line 759: Is this already captured in the commit message as observable behavior? If yes resolve otherwise propose update to the relevant section

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudPreparedStatement.java
```diff
@@ -51,25 +53,29 @@
...
     private String sql;
-    private final ParameterManager parameterManager;
+    private final ParameterAccumulator parameterManager;
     private final Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
     // True if we are currently fetching metadata from the server, this influences the query param generation
     // to not return any data.
     private boolean fetchingMetadata = false;
 
-    DataCloudPreparedStatement(DataCloudConnection connection, ParameterManager parameterManager) {
+    DataCloudPreparedStatement(DataCloudConnection connection, ParameterAccumulator parameterManager) {
```
mkaufmann: Why is the accumulator passed as a parameter and not just constructed internally? Here and other constructor

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudPreparedStatement.java
```diff
@@ -131,52 +141,61 @@ public int executeUpdate() throws SQLException {
...
     @Override
     public void setDouble(int parameterIndex, double x) throws SQLException {
-        setParameter(parameterIndex, Types.DOUBLE, x);
+        setParameter(parameterIndex, HyperType.float8(false), x);
     }
 
     @Override
     public void setBigDecimal(int parameterIndex, BigDecimal x) throws SQLException {
-        setParameter(parameterIndex, Types.DECIMAL, x);
+        HyperType type = x != null ? HyperType.decimal(x.precision(), x.scale(), true) : HyperType.decimal(0, 0, true);
```
mkaufmann: This is hard to read

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/ArrowUtils.java
```diff
@@ -0,0 +1,98 @@
...
+                && type.getPrecision() <= 0
+                && parameterBinding.getValue() instanceof BigDecimal) {
+            BigDecimal bd = (BigDecimal) parameterBinding.getValue();
+            return HyperType.decimal(bd.precision(), bd.scale(), type.isNullable());
+        }
+        return type;
+    }
+
+    public static byte[] toArrowByteArray(List<ParameterBinding> parameters, Calendar calendar) throws IOException {
+        RootAllocator allocator = new RootAllocator(Long.MAX_VALUE);
```
mkaufmann: Shouldn't this also be wrapped in a try with resource? Also check other places in the code base where potentially a RootAllocator might be leaked

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/ColumnMetadata.java
```diff
@@ -0,0 +1,38 @@
...
+public class ColumnMetadata {
+    String name;
+    HyperType type;
+    /**
+     * Optional override for {@code ResultSetMetaData.getColumnTypeName}; {@code null} means the
+     * JDBC layer should use the default derived from {@link #type}. Used for metadata result sets
+     * (e.g. {@code DatabaseMetaData.getTables}) where the JDBC spec pins a specific column-type
+     * label that differs from the underlying Hyper type.
+     */
+    String typeName;
```
mkaufmann, line 33: This I don't understand can you add an example in the comment?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/HyperTypeToArrow.java
```diff
@@ -0,0 +1,100 @@
...
+            case DECIMAL:
+                // Arrow's Decimal requires a positive precision. For under-specified DECIMAL
+                // parameters (bound without precision/scale), callers should derive precision
+                // and scale from the actual BigDecimal value before calling here.
+                if (type.getPrecision() <= 0) {
+                    throw new IllegalArgumentException(
+                            "DECIMAL HyperType must carry precision before Arrow conversion");
+                }
+                return new ArrowType.Decimal(type.getPrecision(), type.getScale(), 128);
+            case CHAR:
```
mkaufmann, line 109: Please review this reverse mapping again. IMHO there should be a test that if one goes from HyperType to Arrow and then uses ArrowToHyper again that the results match exactly across all type properties. Please add such consistency check. If there are cases where behavior can be improved in HyperTypeToArrow to ensure matching adjust the code. If there are cases where you think it doesn't make sense mark those deviations clearly int he test case with documentation

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/hyper/DatabaseMetadataIntegrationTest.java
```diff
@@ -258,30 +247,25 @@ void getColumns_bigint() {
...
     @SneakyThrows
     void getColumns_numeric_10_5_exposesHardcodedScale() {
         val info = getColumnInfo("col_numeric_10_5");
-        assertThat(info.get("TYPE_NAME")).isEqualTo("NUMERIC");
-        assertThat(info.get("DATA_TYPE")).isEqualTo(Types.NUMERIC);
+        assertThat(info.get("TYPE_NAME")).isEqualTo("DECIMAL");
+        assertThat(info.get("DATA_TYPE")).isEqualTo(Types.DECIMAL);
         // BUG: DECIMAL_DIGITS is hardcoded to 2 regardless of the actual scale. The SQL template
         // fetches atttypmod but QueryMetadataUtil ignores it and returns the constant 2.
         // Expected for numeric(10,5): DECIMAL_DIGITS=5.
```
mkaufmann: Can this be fixed?

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/hyper/DatabaseMetadataIntegrationTest.java
```diff
@@ -302,11 +286,8 @@ void getColumns_text() {
...
     void getColumns_varchar255() {
         val info = getColumnInfo("col_varchar_255");
-        // BUG: format_type() returns "character varying(255)" with a length modifier; the map
-        // has no prefix/contains handling for it, so no mapping happens.
-        // Expected: TYPE_NAME="VARCHAR", DATA_TYPE=Types.VARCHAR (12), COLUMN_SIZE=255.
-        assertThat(info.get("TYPE_NAME")).isEqualTo("character varying(255)");
-        assertThat(info.get("DATA_TYPE")).isEqualTo(1043); // pg typoid varchar — NOT Types.VARCHAR
+        assertThat(info.get("TYPE_NAME")).isEqualTo("VARCHAR");
+        assertThat(info.get("DATA_TYPE")).isEqualTo(Types.VARCHAR);
         // BUG: COLUMN_SIZE is hardcoded to 255 in QueryMetadataUtil (accidentally "correct" here).
```
mkaufmann: Can this be fixed?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/resultset/SimpleResultSet.java
mkaufmann: Applied. I extracted the switch logic into `HyperTypes.isIntegerLike(HyperType)` / `isStringLike(HyperType)` helpers whose own switches are exhaustive (no `default`), and used the helpers in `SimpleResultSet`. That way a new `HyperTypeKind` forces a compile-time decision in one place rather than sprinkling every switch in the codebase, but the lint signal is still there.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/DefaultParameterAccumulator.java
mkaufmann: Flattened. `ParameterAccumulator` is now the single concrete class (no interface, no `DefaultParameterAccumulator`) since there's only one implementation and we don't need the seam for mocking anymore — the prepared-statement test verifies against the real accumulator's `getParameters()` list now.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
mkaufmann: Keeping it at BIGINT — the reference (`jdbc-reference/src/main/resources/reference.json`, captured via pgjdbc) already surfaces `oid` as `Types.BIGINT` (-5), and `HyperType.oid` exists precisely so the pg_catalog path can preserve the distinction internally (precision=10 for the digit count, display size, etc.) while still matching what pgjdbc clients expect at the JDBC surface.

Switching to `OTHER` would:
1. Break reference-parity (JDBCReferenceTest already skips `oid` via `isTestable=false` because Hyper and Postgres disagree on semantics — but the JDBC-surface mapping they do agree on).
2. Make `getLong("oid_col")` stop working for callers that treat it as an integer.

I think the right refinement, if any, is to narrow the internal `HyperType.oid` precision/display-size behavior rather than change the JDBC code. Happy to revisit if you want a different stance.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
mkaufmann: Both are real:
- `HyperTypeKind.VARBINARY` is Hyper's `bytea`. Arrow produces it via `ArrowType.Binary` → `HyperType.varbinary` in `ArrowToHyperTypeMapper.visit(ArrowType.Binary)`.
- `HyperTypeKind.BINARY` (fixed-length) is produced by `ArrowToHyperTypeMapper.visit(FixedSizeBinary)` — Hyper doesn't have a SQL-level fixed-length binary type, but its Arrow response schema can carry `FixedSizeBinary` (e.g. for some hash types, UUID-like fixed blobs), so we need to map it on the way in.

Dropped the BINARY branch would make those schemas fall through to the `UNKNOWN` escape hatch, which is worse than having a typed accessor. I added a round-trip test (`HyperTypeArrowRoundtripTest`) that pins this exact encoding so if someone considers removing it in the future, the test tells them what breaks.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
mkaufmann: Dropped. The `default: throw` moved out of the switch — unmatched JDBC codes fall through to the post-switch `throw new IllegalArgumentException`. (Since `fromJdbcTypeCode` switches on an `int`, the lint doesn't fire the same way as for enum switches, but the shape is consistent with your stated preference.)

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
mkaufmann: Agreed — kept this one. `toJdbcTypeName`'s `default` is the shared "use `toJdbcType(t).getName()`" fallback, not an error path.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
mkaufmann: Verified against `reference.json` (pgjdbc baseline):
```
select NULL::real          → float4 scale=8  prec=8
select double precision '0.5' → float8 scale=17 prec=17
```
pgjdbc does indeed report `scale == precision` for binary floats (8 for REAL, 17 for DOUBLE). Rewrote the comment to state that explicitly instead of the vague "legacy behaviour".

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
mkaufmann: No — the prototype's scale field doesn't flow into any observable column of `getTypeInfo()`. For DECIMAL the row gets:
- `PRECISION` = `prototype.getPrecision()` = **38** (max precision supported — right).
- `MINIMUM_SCALE` = 0, `MAXIMUM_SCALE` = 38 — set separately at `HyperTypes.java:528-536` based on the kind, not from `prototype.getScale()`.
- `toHyperTypeName(prototype)` would render `"numeric(38,0)"` only if we were emitting it; but the row uses `LOCAL_TYPE_NAME = prototype.toHyperTypeName()` which reads the prototype's scale of 0. That's visible (`getTypeInfo()` shows `LOCAL_TYPE_NAME = "numeric(38,0)"`). Not ideal — a client asking for the generic DECIMAL typename would prefer just `"numeric"`.

I'll leave the prototype as-is for this PR but filed the `LOCAL_TYPE_NAME` nit mentally for a follow-up; the simplest fix is a special case in the DECIMAL row construction rather than in the prototype.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudDatabaseMetadata.java
mkaufmann: Yes — captured in the commit message, second-to-last paragraph of the "Migrated callers" block: *"DataCloudDatabaseMetadata.getTypeInfo() returns a real 18-column result derived from HyperTypeKind (was empty before); UNKNOWN and NULL are excluded from the enumeration"*. Happy to move it into the "Observable behaviour changes" list for more visibility if you prefer — let me know.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudPreparedStatement.java
mkaufmann: Done. Both constructors no longer take a `ParameterAccumulator`; it's constructed internally as a `final` field. The JDBC `prepareStatement` caller just does `new DataCloudPreparedStatement(this, sql)`. Tests that previously mocked the accumulator now assert against `preparedStatement.parameters.getParameters()` (package-private field accessor).

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudPreparedStatement.java
mkaufmann: Rewrote as an explicit if/else with a comment:
```java
final HyperType type;
if (x == null) {
    // Precision/scale are unknown; let Arrow encoding pick a suitable default.
    type = HyperType.decimal(0, 0, true);
} else {
    type = HyperType.decimal(x.precision(), x.scale(), true);
}
setParameter(parameterIndex, type, x);
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/ArrowUtils.java
mkaufmann: Wrapped `toArrowByteArray`'s allocator in try-with-resources alongside the `VectorSchemaRoot`.

Audited the rest of the codebase:
- `QueryResultArrowStream.asArrowStreamReader` (line 50) creates `new RootAllocator(...)` and hands it to `ArrowStreamReader` without a try-with-resources. Fixing this needs the `ArrowStreamReader` lifecycle to propagate the close to the allocator — that's a bigger change than worth bundling into this PR since the consumer may expect the allocator to outlive the reader. I'd rather not rush it.

I'll open a focused follow-up for the stream-reader allocator. Flagged it here so it's not lost.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/ColumnMetadata.java
mkaufmann: Added a concrete example to the javadoc:

> Example: the columns of `DatabaseMetaData.getTables` are spec'd as `TEXT` and `SHORT` (see JDBC 4.2 §28.12), not `VARCHAR` / `SMALLINT`. `MetadataSchemas` constructs those columns as `new ColumnMetadata("TABLE_NAME", HyperType.varcharUnlimited(true), "TEXT")` so that `ResultSetMetaData.getColumnTypeName` returns `"TEXT"` while the accessor machinery still sees a `VARCHAR`. Regular query result columns pass `null` (via the two-arg constructor) and get `"VARCHAR"` from `HyperType`.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/HyperTypeToArrow.java
mkaufmann: Added `HyperTypeArrowRoundtripTest`. It covers:

- **Positive round-trips**: parameterized test that `HyperType → ArrowType → HyperType` reproduces the original for every kind we can encode (BOOL, INT*, OID, FLOAT*, DECIMAL, VARCHAR unlimited, BINARY, VARBINARY, DATE, TIME, TIMESTAMP, TIMESTAMP_TZ, ARRAY of int/varchar).

- **Non-encodable kinds** (`nonEncodableKinds()`): INTERVAL, JSON, UNKNOWN, NULL — documented why each can't encode.

- **Known lossy cases** (`lossyRoundtrip_*` tests): three real divergences the test flushed out —
  1. `CHAR(n)` → Utf8 → `VARCHAR unlimited` (Arrow has no fixed-length string)
  2. `VARCHAR(n)` → Utf8 → `VARCHAR unlimited` (no length metadata stamped)
  3. `TIME_TZ` → `Time(MICROSECOND, 64)` → `TIME` (Arrow Time has no timezone slot)

Each is pinned with a comment explaining the root cause. Fixing any of them needs `HyperTypeToArrow` to stamp `hyper:*` metadata on the Arrow `Field` and `ArrowToHyperTypeMapper` to read it back — that's legitimate follow-up work rather than silently landing now.

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/hyper/DatabaseMetadataIntegrationTest.java
mkaufmann: Fixed. `QueryMetadataUtil.constructColumnData` now forwards `HyperTypes.getPrecision(hyperType)` as `COLUMN_SIZE` (pulled from the `HyperType.precision` field that `PgCatalogTypeParser` extracted from `format_type(atttypmod)`), and uses `hyperType.getScale()` for `DECIMAL_DIGITS` on DECIMAL columns. So `numeric(10,5)` now reports `COLUMN_SIZE=10, DECIMAL_DIGITS=5` and `char(1)` reports `COLUMN_SIZE=1`. Updated the test to assert the fixed values (dropped the `BUG:` comment).

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/hyper/DatabaseMetadataIntegrationTest.java
mkaufmann: Same fix as above — `COLUMN_SIZE` for `varchar(255)` still happens to be 255, but it's now derived from the HyperType rather than hardcoded, and the BUG comment is gone. The cross-check `forPrecision` test now asserts the two paths are in full agreement (mismatch set empty).

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/resultset/SimpleResultSet.java
```diff
@@ -288,20 +272,15 @@ public Array getArray(int columnIndex) throws SQLException {
...
+            val v = getInt(columnIndex);
+            return wasNull ? null : v;
         }
-        throw new SQLException("Unsupported column type in `getObject`: "
-                + metadata.getColumn(columnIndex).getType().toString());
+        if (HyperTypes.isStringLike(type)) {
+            return getString(columnIndex);
+        }
+        throw new SQLException("Unsupported column type in `getObject`: " + type);
     }
```
mkaufmann, line 294: Why is this so much special cased to INT32 what about all the other types?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
mkaufmann: Ok that makes snse maybe add a comment here to explain that

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudDatabaseMetadata.java
mkaufmann: Yes move it as allowing to enumerate types is a nice addition

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/HyperTypeToArrow.java
mkaufmann: >Each is pinned with a comment explaining the root cause. Fixing any of them needs HyperTypeToArrow to stamp hyper:* metadata on the Arrow Field and ArrowToHyperTypeMapper to read it back — that's legitimate follow-up work rather than silently landing now.

Please attempt to land that now so that we have a clean slate

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/resultset/SimpleResultSet.java
mkaufmann: Good catch — this class is the base for the JDBC metadata result sets (the ones `DatabaseMetaData.getTables/Columns/etc.` return), so the `HyperTypeKind` universe is narrower than full query results. The INT32 special case is there because metadata columns like `DATA_TYPE` and `NULLABLE` are `INTEGER`. For a plain query-result `ResultSet` the streaming path (`StreamingResultSet`) is used and covers the full kind set.

Rather than broaden these switches to cover every kind (which would spread the "one-switch-per-column-family" pattern across four methods), I added a class-level doc note flagging `SimpleResultSet` for removal: the plan is to migrate `DataCloudMetadataResultSet` onto `StreamingResultSet` so there's only one result-set implementation. Tracked as follow-up. Commit: d823847.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/types/HyperTypes.java
mkaufmann: Added an inline comment on the OID case in `HyperTypes.toJdbcType`:

> // Hyper's OID is an unsigned 32-bit internal type. We surface it as BIGINT to
> // match the pgjdbc reference (see JDBCReferenceTest): pgjdbc reports oid with
> // Types.BIGINT, which keeps \`getLong("oid_col")\` working for integer-treating
> // callers. HyperType.oid stays a distinct kind so the pg_catalog path can keep
> // precision/display accurate (10 digits for the u32 range).

Commit: d823847.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudDatabaseMetadata.java
mkaufmann: Called it out in the layered commit message (d823847). I'm not amending `c647bce`'s message to avoid rewriting a published commit on a branch with an open PR, per your earlier guidance — the `getTypeInfo` behavior change is noted alongside the other observable ones in the d823847 body and will end up in the final squashed message when the PR is merged.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/HyperTypeToArrow.java
mkaufmann: Done for CHAR / CHAR(1) / VARCHAR(n); TIME_TZ stays lossy with a better rationale. Commit: d823847.

**Fixed by stamping the same `hyper:*` metadata Hyper's server already emits:**
- `hyper:type = "Char1"` for CHAR(1)
- `hyper:type = "Char"` + `hyper:max_string_length = <n>` for CHAR(n)
- `hyper:max_string_length = <n>` alone for bounded VARCHAR(n)

`ArrowToHyperTypeMapper` already reads these keys (it was wired up for Hyper's inbound Arrow stream), so the decode path needed no changes. Moved CHAR(1), CHAR(16), VARCHAR(255), VARCHAR(1024) into the positive parameterised cases of `HyperTypeArrowRoundtripTest`; the unbounded-VARCHAR case also round-trips.

**TIME_TZ stays pinned as lossy** — rewrote the comment to explain why it's not a simple metadata stamp:

1. Arrow's `Time` has no timezone slot (unlike `Timestamp.timezone`), so the offset can't ride on the ArrowType.
2. SQL `time with time zone` stores a **per-value** UTC offset — a single column can contain `14:00+02` and `14:00-05`. Field-level metadata is per-column, so it can't carry a value-varying offset.
3. Our read-side `QueryJDBCAccessorFactory` has no TIME_TZ branch either — server-returned `timetz` columns already collapse to `TimeVectorAccessor`, dropping the offset on ingress. Fixing only the write side would make the asymmetry worse.

Proper round-trip needs either a synthetic Arrow schema (e.g. `struct{time, offset_seconds}`) or Hyper-protocol coordination. Legitimate follow-up rather than silent assumption; pinned with the new comment so a future contributor sees the full reasoning.

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/protocol/data/HyperTypeArrowRoundtripTest.java
```diff
@@ -96,37 +96,27 @@ void nonEncodableKinds() {
...
+        //     "14:00-05" in the same column). Field metadata is per-column, so it cannot carry
+        //     a value-varying offset.
+        //   - The driver's read-side QueryJDBCAccessorFactory has no TIME_TZ branch either —
+        //     server-returned `timetz` columns already collapse to the plain TimeVectorAccessor,
+        //     so the offset is dropped on ingress as well. Fixing the write-side alone would
+        //     create an asymmetry worse than the current uniform loss.
+        //
+        // Proper round-trip needs a synthetic schema (e.g. struct{time, offset_seconds}) or a
+        // Hyper-protocol coordination. Leaving TIME_TZ explicitly pinned as lossy until that
+        // design lands.
```
mkaufmann: Woulnd't formulate as bug but that it looks like it's intended behavior that the TZ part is only important when passing it in but not when returning

mkaufmann: Good framing — you're right, "lossy" is the wrong label. Reframed the comment as intended asymmetry: the offset is honored on input (the value is interpreted in the supplied offset), but by design it isn't preserved on return, matching what Hyper's read side already does (server normalises, accessor emits time-of-day without offset). Renamed the test from `lossyRoundtrip_timeTzCollapsesToTime` to `timeTzAsymmetry_preservedOnInputButDroppedOnOutput` and rewrote the rationale block. Commit: 127b6ac.

j10t (approved):

KaviarasuSakthivadivel (approved):

PR #175, opened 2026-04-24
refactor: unify ResultSet implementations on Arrow-backed path

mkaufmann:
## Summary

Collapse the two ResultSet families (streaming Arrow + row-based metadata) onto a single Arrow-backed pipeline so there is one accessor implementation, one set of type semantics, and one place to fix bugs. Tighten root-allocator hygiene end-to-end while we are in there, and bring `getTypeInfo()` and integer-column accessor coercion into line with JDBC 4.2.

## Why

The driver previously carried two parallel result-set implementations: `StreamingResultSet` for query results (Arrow IPC, columnar accessors) and `SimpleResultSet` / `DataCloudMetadataResultSet` / `ColumnAccessor` for metadata (row-oriented `List<List<Object>>`, hand-rolled per-cell coercion). Same JDBC surface, two divergent code paths. Bugs found in one were rarely fixed in the other; type semantics drifted (e.g. `getBoolean` on an integer column behaved differently between the two), and the metadata path silently `toString()`'d any payload you handed it. The pre-rebase review of this PR also surfaced several allocator leak windows and JDBC-spec compliance gaps that the unification made it natural to fix.

## What changed

**Unified result set.** `StreamingResultSet`, `DataCloudMetadataResultSet`, `SimpleResultSet`, and `ColumnAccessor` are removed. Every JDBC metadata call (`getTables`, `getColumns`, `getSchemas`, `getTypeInfo`, the empty-metadata helpers) now flows through `DataCloudResultSet` via a new `MetadataResultSets` factory. `MetadataResultSets` builds a single-batch Arrow IPC stream by reusing `VectorPopulator` (the same code path the JDBC parameter encoder uses) and `HyperTypeToArrow.toField`, then hands the resulting reader + allocator to `DataCloudResultSet.of`.

`DataCloudResultSet` is now a `public class` rather than the prior empty marker interface; the concrete implementation is no longer a sibling type called `StreamingResultSet`. The `of(...)` factory takes a `QueryResultArrowStream.Result` (reader + allocator pair) and owns both their lifecycles.

**Root allocator hygiene.** Six independent leak windows closed:
- `QueryResultArrowStream.toArrowStreamReader` returns a `Result` holder that pairs reader + allocator and closes both in order (reader first so ArrowBuf accounting clears before the allocator's budget check). The 100 MB cap moves to a public `ROOT_ALLOCATOR_BUDGET_BYTES` constant and is now reused by the metadata path.
- `MetadataResultSets.of` and `QueryResultArrowStream.toArrowStreamReader` both wrap allocator + reader construction in try/catch so the allocator is closed if `new ArrowStreamReader(...)` ever throws before ownership transfers.
- `DataCloudResultSet.of`'s construction-failure cleanup now wraps both reader.close and allocator.close with `addSuppressed` so neither close masks the original `SQLException`.
- `ArrowStreamReaderCursor.close` uses try-with-resources over `(allocator, reader)` so reader closes first and any allocator-close exception attaches as suppressed onto the reader's instead of replacing it.
- `DataCloudStatement.executeQuery` and `getResultSet` hoist `iterator.getQueryStatus().getQueryId()` once before allocator construction, so a throw between allocator creation and `DataCloudResultSet.of` taking ownership can no longer strand the allocator.
- `DataCloudResultSet.close` is idempotent across cursor.close failures: the `closed` flag flips before delegating, so a defensive caller's retry is a no-op rather than a double-close.

**JDBC spec compliance.**
- `getTypeInfo()` boolean columns (`CASE_SENSITIVE`, `UNSIGNED_ATTRIBUTE`, `FIXED_PREC_SCALE`, `AUTO_INCREMENT`) are now declared as `BOOLEAN` per JDBC 4.2 (DatabaseMetaData.getTypeInfo Javadoc) and pgjdbc, via a new `bool(...)` helper in `MetadataSchemas`. They were previously declared as `text(...)` while the row producer wrote raw `Boolean` values, which only "worked" because `VarCharVectorSetter` silently `toString()`'d everything.
- `BaseIntVectorAccessor.getBoolean` now matches `ResultSet.getBoolean`'s spec text on integer columns: 0 returns false, 1 returns true, anything else throws `SQLException` with SQLState `22018` (matching pgjdbc's strict CANNOT_COERCE behavior in `BooleanTypeUtil.fromNumber`). Previously inherited the abstract default that threw `SQLFeatureNotSupportedException` for everything.
- `VectorPopulator.VarCharVectorSetter` is tightened from `<VarCharVector, Object>` to `<VarCharVector, String>`, so non-String payloads fail fast at the `BaseVectorSetter` type guard. The `byte[]` arm was dead — `setBytes` / `setBinaryStream` / `setUnicodeStream` / `setAsciiStream` all throw FEATURE_NOT_SUPPORTED in `DataCloudPreparedStatement`.
- Integer-family setters (`IntVectorSetter`, `SmallIntVectorSetter`, `TinyIntVectorSetter`) now range-check before narrowing rather than silently truncating; `DecimalVectorSetter` does the same via a `bitLength() > 63` guard on the unscaled value.
- `QueryJDBCAccessor.getObject(Class)` gains an `isInstance` fallback so `getObject(col, String.class)` on a VARCHAR (and analogous identity-class paths on every other accessor) works without each accessor implementing typed `getObject` itself.

## Observable behavior changes

- `rs.getBoolean("CASE_SENSITIVE")` on `getTypeInfo()` returns a real Boolean (was `SQLFeatureNotSupportedException` via the broken VARCHAR path).
- `rs.getBoolean("NULLABLE")` on `getColumns()` (and any other integer column) returns `false` for `0` and `true` for `1`, instead of throwing. Other integer values throw `SQLException` (SQLState `22018`).
- `rs.getDate(intCol)` / `getTime(intCol)` / `getTimestamp(intCol)` on metadata rows throw `SQLException` (was `UnsupportedOperationException`).
- `rs.getObject(intCol, Boolean.class)` on metadata rows now throws (the strict `isInstance` path).
- `rs.getMetaData().getColumnType(...)` on the four `getTypeInfo()` boolean columns returns `Types.BOOLEAN`, not `Types.VARCHAR`.
- `rs.getMetaData().getColumnTypeName(...)` on every metadata result set (`getTables`, `getColumns`, `getTypeInfo`, …) returns the JDBC type name derived from the column's `HyperType` (`"VARCHAR"`, `"SMALLINT"`, `"INTEGER"`, `"BOOLEAN"`) rather than the prior Hyper-flavored labels (`"TEXT"`, `"SHORT"`, `"INTEGER"`, `"BOOL"`). The JDBC spec only requires *some* database-specific type name and does not pin specific strings; this aligns with the names other JDBC consumers in the driver already use.
- `ps.setObject(idx, x, Types.VARCHAR)` with a non-String / non-byte[] argument now throws `IllegalArgumentException` instead of silently `toString()`-ing the argument.
- `ps.setObject(idx, x, Types.INTEGER)` (and INT2/INT8) throws `IllegalArgumentException` for out-of-range Numbers instead of silently narrowing; same for `Types.DECIMAL` when the unscaled value exceeds 64 bits.

## Breaking changes

`com.salesforce.datacloud.jdbc.core.DataCloudResultSet` is now a `public class` rather than a `public interface`. External code that wrote `class MyRs implements DataCloudResultSet` (decorators, wrappers, hand-rolled doubles) no longer compiles; code that consumes the standard `java.sql.ResultSet` / `DataCloudResultSet` API as an opaque type recompiles unchanged.

The previously-public types `StreamingResultSet`, `DataCloudMetadataResultSet`, `SimpleResultSet`, `ColumnAccessor` are removed. External callers of `StreamingResultSet.of(ArrowStreamReader, ...)` should switch to `DataCloudResultSet.of(QueryResultArrowStream.Result, ...)`.

## Test plan

- [x] `./gradlew :jdbc-core:test` — full module suite green.
- [x] `./gradlew :jdbc-core:spotlessCheck` — formatting clean.
- [x] `./gradlew clean build` — full build including `:spark-datasource`, JaCoCo coverage, verification.
- New tests pin the load-bearing invariants: `MetadataSchemasTest` adds three TYPE_INFO position-by-position assertions; `VarCharVectorSetterStrictTypeTest` regresses on Boolean / byte[] / Number payloads; `IntegerVectorSetterRangeCheckTest` extends to `DecimalVectorSetter`; `ArrowStreamReaderCursorTest` pins reader-before-allocator close ordering plus `addSuppressed` chaining when both throw; `DataCloudResultSetMethodTest` pins `close()` idempotence under cursor.close failure; `DataCloudDatabaseMetadataTest.testGetTypeInfo` now exercises `getBoolean` on all four boolean columns end-to-end.

BREAKING CHANGE: `DataCloudResultSet` is now a class instead of an interface; `StreamingResultSet`, `DataCloudMetadataResultSet`, `SimpleResultSet`, `ColumnAccessor` are removed; metadata int-column `getDate`/`getTime`/`getTimestamp` throw `SQLException` (was `UnsupportedOperationException`); `getTypeInfo()` boolean columns are typed `BOOLEAN` instead of `VARCHAR` (`getObject` returns `Boolean`, not `String`); `getColumnTypeName` on metadata result sets returns the JDBC type name (`VARCHAR`/`SMALLINT`/`INTEGER`/`BOOLEAN`) instead of the prior Hyper-flavored labels (`TEXT`/`SHORT`/`INTEGER`/`BOOL`); `ps.setObject` with `Types.VARCHAR` rejects non-String/byte[] payloads; integer-family and DECIMAL setters reject out-of-range values instead of silently narrowing.


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/MetadataArrowBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+     * <p>Values are coerced to the target vector type on a best-effort basis (e.g. a
+     * {@link Boolean} in a VARCHAR column becomes {@code "true"}/{@code "false"}). This mirrors
+     * the loose coercion the removed {@code SimpleResultSet.getString} used to do on row-based
+     * metadata data.
+     */
+    public static Result build(List<ColumnMetadata> columns, List<List<Object>> rows) {
+        RootAllocator allocator = new RootAllocator(Long.MAX_VALUE);
+        VectorSchemaRoot root = null;
+        try {
+            Schema schema = buildSchema(columns);
```
mkaufmann: Isn't this use case very similar to the parameter encoding (except that the kind of only has one row while here we need multiple rows)? Can we use the same builder approach for both? I guess the input could be List<ColumnMetadata> columns, List<List<Object>> rows in both cases but in the parameter case it's only a single row

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
```diff
@@ -21,31 +20,79 @@
...
+     * Create a cursor that pulls batches from an {@link ArrowStreamReader}. The reader (and the
+     * allocator it was constructed with) are owned by the cursor — closing the cursor closes the
+     * supplied {@code ownedResources}.
+     */
     @SneakyThrows
-    private VectorSchemaRoot getSchemaRoot() {
-        return reader.getVectorSchemaRoot();
+    static ArrowStreamReaderCursor streaming(
+            ArrowStreamReader reader, AutoCloseable ownedResources, ZoneId sessionZone) {
+        val root = reader.getVectorSchemaRoot();
```
mkaufmann: Instead of adding so many new variants can't we keep the old interface and instead provide the metadta rows through an arrowstreamreader that is backed by memory buffers?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/MetadataArrowBuilder.java
mkaufmann: Good call. Fixed in 92929bf — both paths now go through VectorPopulator.populateVectors: the parameter-binding path (single row) and the metadata path (many rows) both dispatch through the same setCell primitive. Each setter is now parameterised by row index. MetadataArrowBuilder is deleted; metadata results are built by the same code as parameter binding, wrapped in an ArrowStreamReader over a ByteArrayInputStream so they land on the normal StreamingResultSet path.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
mkaufmann: Agreed. Fixed in 92929bf — ArrowStreamReaderCursor is back to its original ArrowStreamReader-based interface (plus an owned BufferAllocator for lifecycle). Metadata results now serialise a populated VectorSchemaRoot to Arrow IPC bytes and wrap them in a ByteArrayInputStream-backed ArrowStreamReader, so there are no inMemory/streaming variants: every ResultSet flows through the same reader/cursor plumbing. StreamingResultSet.ofInMemory is gone. The JDBC type-name override (e.g. "TEXT") rides along via a jdbc:type_name field-metadata key instead of a sidechannel parameter.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -73,36 +73,40 @@ private StreamingResultSet(
...
     }
 
     /**
-     * Creates a StreamingResultSet with a specified session timezone.
+     * Creates a StreamingResultSet from an {@link ArrowStreamReader} and its backing allocator.
+     *
+     * <p>Ownership of both the reader and the allocator transfers to the returned result set —
+     * closing the result set closes the reader and then the allocator, in that order, so Arrow's
+     * buffer accounting clears before the allocator's budget check. Callers must not close
+     * either separately.
```
mkaufmann: As an alternative idea, would it make sense to have the allocator hosted at the connection? Would that bring performance benefits and have that allocator reused and only closed at connection close time?

mkaufmann: Worth thinking through. Here is my take:

**Performance:** minor, if anything. A RootAllocator is just accounting state — cheap to construct. The actual memory cost is in the ArrowBufs, which is per-result-set either way. I would not expect a measurable win.

**Observability:** real win. A connection-wide root gives a single place to see total memory under one connection, and lets us cap per-connection memory globally instead of per result set (which is the JDBC-correct unit anyway).

**Lifecycle:** the standard Arrow pattern would be `connection.rootAllocator` → per-result-set child allocator via `newChildAllocator`. Closing a child returns its budget to the parent; closing the parent on `connection.close()` forcibly invalidates any still-open children. That is actually closer to JDBC spec behavior (connection close invalidates statements/result sets) than what we do today.

**Bug isolation:** slight regression. Today a leaking RS cannot affect another RS. With child allocators off one root, a leak blocks the whole connection from closing cleanly. In practice the leak detector still names the child so diagnosis is fine.

**Scope:** non-trivial change: need to propagate the connection allocator through `QueryResultArrowStream.toArrowStreamReader`, `SQLExceptionQueryResultIterator.createSqlExceptionArrowStreamReader`, and the metadata path. Each RS would take a child allocator and close it on RS close; connection close closes the root.

I would like to do it, but lean toward landing this PR as-is (unified code paths + current per-RS allocators) and doing the connection-hosted allocator as a follow-up. That keeps this PR focused on "unify the two result sets" and gives the allocator-refactor its own diff to review. Happy to do it here if you prefer.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -329,10 +360,32 @@ public Object getObject(int columnIndex, Map<String, Class<?>> map) throws SQLEx
...
+                return null;
+            }
+            if (type.isInstance(raw)) {
+                return type.cast(raw);
+            }
+            throw new SQLException(
+                    "Cannot convert column value to " + type.getName() + "; actual type is "
+                            + raw.getClass().getName(),
+                    ex);
+        }
```
mkaufmann: Isn't this putting a bandaid over bad behavior of the accessors? Shouldn't the accessor already implement this behavior so that we don't need the bandaid here? If true check if if this proper fix can be factored into a separate commit (before the others on this branch) so that I can maybe ship it independently

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -73,36 +74,40 @@ private StreamingResultSet(
...
+     * Creates a StreamingResultSet from an {@link ArrowStreamReader} and its backing allocator.
+     *
+     * <p>Ownership of both the reader and the allocator transfers to the returned result set —
+     * closing the result set closes the reader and then the allocator, in that order, so Arrow's
+     * buffer accounting clears before the allocator's budget check. Callers must not close
+     * either separately.
+     *
+     * <p>The column metadata (including any {@link ColumnMetadata#getTypeName()} override
+     * stamped under {@link com.salesforce.datacloud.jdbc.protocol.data.HyperTypeToArrow#JDBC_TYPE_NAME_METADATA_KEY})
+     * is derived from the Arrow schema via {@link ArrowToHyperTypeMapper#toColumnMetadata(org.apache.arrow.vector.types.pojo.Field)}.
```
mkaufmann: This is too low level comment, remove

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
```diff
@@ -112,6 +117,28 @@ public static StreamingResultSet of(ArrowStreamReader resultStream, String query
...
+            try {
+                arrowStream.getReader().close();
+            } catch (Exception suppressed) {
+                ex.addSuppressed(suppressed);
+            }
+            arrowStream.getAllocator().close();
+            throw ex;
+        }
+    }
+
```
mkaufmann: This looks like an independent fix, if yes factor (and the tests of course) into a seperate commit at the start of the branch so that I can potentially ship it independently

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
```diff
@@ -54,10 +66,18 @@ private QueryJDBCAccessor createAccessor(FieldVector vector) throws SQLException
...
-        if (reader.loadNextBatch()) {
-            currentIndex.set(0);
-            return true;
+    /**
+     * Load the next batch that has at least one row, skipping any zero-row batches in between.
+     * Empty IPC batches are valid per the Arrow spec and can be emitted by upstream producers
+     * (long-running queries returning empty initial chunks, schema-only metadata streams), so
+     * "successfully loaded" must not be conflated with "row available."
+     */
+    private boolean loadNextNonEmptyBatch() throws IOException {
```
mkaufmann, line 72: Why do we need to skip empty IPC batches if they are valid? Shouldn't callers handle them correctly?

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/metadata/MetadataResultSetsTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @Test
+    void getType() throws SQLException {
+        assertThat(emptyMetadataResultSet.getType()).isEqualTo(ResultSet.TYPE_FORWARD_ONLY);
+    }
+
+    @Test
+    void getFetchDirection() throws SQLException {
+        assertThat(emptyMetadataResultSet.getFetchDirection()).isEqualTo(ResultSet.FETCH_FORWARD);
+    }
```
mkaufmann: These are superflous remove

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
mkaufmann: Agreed — fixed in 322663e by giving `QueryJDBCAccessor.getObject(Class)` a default `raw + isInstance` implementation, so accessors that don't need typed conversion just inherit the right behavior. Then a6da614 collapses `StreamingResultSet.getObject(int, Class)` to a direct dispatch (no try/catch).

Per your "factor into a separate commit at the start of the branch" ask: 322663e is now the **first commit** on the branch (touches only `QueryJDBCAccessor.java`, applies cleanly on `main`), so it can be cherry-picked and shipped independently of the rest of this PR. The follow-up that drops the bandaid (a6da614) sits at the end since it depends on the rest of the unified `StreamingResultSet`.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
mkaufmann: Removed in e329860 — the paragraph that mentioned `JDBC_TYPE_NAME_METADATA_KEY` is gone; the docstring now stays at the level of "this is what `of(reader, allocator, queryId, sessionZone)` does" and leaves the field-metadata details documented at `HyperTypeToArrow` / `ColumnMetadata` where they live.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
mkaufmann: Looked at this carefully — the fix is already in its own commit (26bdc37 `fix: close allocator on StreamingResultSet.of failure in query path` with its regression test in `StreamingResultSetMethodTest`), but I left it where it is on the branch rather than moving it to the very front. The reason: as coded, this fix is **not** independent of the unify refactor. Specifically, it adds `StreamingResultSet.ofClosingOnFailure(QueryResultArrowStream.Result, ...)` and switches the four call sites to it — and on `main`:

- `QueryResultArrowStream` doesn't return a `Result` wrapping `(reader, allocator)`; `toArrowStreamReader` just returns an `ArrowStreamReader`.
- `StreamingResultSet.of(...)` only takes `(reader, queryId)` — there's no allocator to leak on construction failure, because the allocator isn't owned by the result set on `main` at all.

So "pull this fix to the front of the branch" would mean rewriting the fix as a different fix against the pre-unify surface, which feels like it deserves its own decision rather than a sneaky reorder here. Two options if you'd like to ship the leak fix independently of this PR:

1. I can carve out a minimal "pre-unify" leak fix against `main` directly (different shape — likely closes the allocator that today lives inside `QueryResultArrowStream` rather than in `StreamingResultSet`) and open it as its own PR. Happy to do this.
2. Land it as part of this PR and treat the unify+leak-fix as a unit.

Lean toward (1) if you want to ship it ahead of this PR, otherwise (2) is fine. Let me know.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
mkaufmann: Fair question. The cursor *is* the caller for a JDBC `ResultSet` — there's no layer above it that has a sensible "I just consumed an empty batch, please advance" hook to expose. Sharpened the docstring on this method (e329860) to explain it directly:

> Empty IPC batches are valid per the Arrow spec, and producers in the wild emit them — Hyper / query-federator can send a schema-only initial chunk, and async chunked-query paths use empty batches as a keep-alive between data chunks. JDBC has no notion of a "batch": `ResultSet.next()` only knows "is there another row?". This cursor is the single place that translates batch-level signals from `ArrowStreamReader.loadNextBatch()` into row-level advances, so it must consume the empty batches itself rather than push that responsibility outward.

Same fix is also independently shipped as commit 4d2e47f (`fix: skip zero-row batches in ArrowStreamReaderCursor`). I left that commit where it is on the branch rather than moving it to the start, because as coded it also removes the now-redundant empty-batch workaround in `MetadataResultSets.writeArrowStream` — which doesn't exist on `main`, so the pre-unify version of the fix would have a different shape (just the cursor side, no metadata-side cleanup). Happy to carve that out if you want a `main`-targeted preview.

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/metadata/MetadataResultSetsTest.java
mkaufmann: Pruned in e329860. The `getRow` / `next` / `isClosed` / `getStatement` / `unwrap` / `isWrapperFor` / `getHoldability` / `getFetchSize` / `setFetchSize` / `getWarnings` / `getConcurrency` / `getType` / `getFetchDirection` slice was testing the `StreamingResultSet` plumbing (which every result set on this branch shares), and is already covered by `StreamingResultSetMethodTest`. Kept the **arity-contract** slice (`shortRowRejected` / `longRowRejected` / `rightArityAccepted` / `nullRowAcceptedAsAllNulls` / `emptyRowsAccepted`) since that *is* metadata-result-set-specific behavior — the row-shape validation lives in `MetadataResultSets.of`, not in the shared streaming plumbing.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
mkaufmann: Can you move this to an independent commit along with test coverage?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
mkaufmann: Try to move it to a separate commit (along with test coverage). Would it make sense to move the QueryResultArrowStream as well then?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
mkaufmann: Done — carved out as commit 2088116 `fix: skip zero-row batches in ArrowStreamReaderCursor`, now the **first commit on the branch** (touches only `ArrowStreamReaderCursor.java` + `ArrowStreamReaderCursorTest.java`, applies cleanly on `main`'s pre-unify cursor signature `(reader, sessionZone)`). It can be cherry-picked and shipped independently.

The downstream commit 049e82d (formerly the same-named "skip zero-row batches" commit) now contains only the residual: removing the now-redundant `MetadataResultSets.writeArrowStream` "skip writeBatch when rowCount==0" workaround, since the cursor handles the empty-only case at the seam.

Test coverage that ships with the pre-unify commit:
- `skipsZeroRowBatchAndYieldsSubsequentNonEmptyRows` — real Arrow IPC stream of {0-row, 1-row} → next() reports the one real row, then false.
- `zeroRowOnlyBatchYieldsNoRows` — real Arrow IPC stream of {0-row} → next() returns false.
- `firstNextReturnsTrueWhenInitialBatchHasRows` / `firstNextReturnsFalseWhenStreamHasNoBatches` — pure-mock pinning of the first-batch logic, replacing the old `forwardsLoadNextBatch(true|false)` parameterised test that would loop forever under the new control flow.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/StreamingResultSet.java
mkaufmann: Honest answer: yes, that would make the leak fix sit on a clean base — but it's a non-trivial split because the unify refactor is what *introduced* the allocator-ownership concept (`Result` holder, cursor's `BufferAllocator` field, `StreamingResultSet.of` taking an allocator). To pull "QueryResultArrowStream introduces `Result`" + "StreamingResultSet/cursor own the allocator" + "ofClosingOnFailure leak fix" forward as a single commit before the unify refactor, I'd be effectively breaking the unify commit into two halves and rebuilding the unify on top of the pre-extracted half. Doable, but it means rewriting a ~800-line refactor.

Given (a) you have one substantive comment left to evaluate (whether commit 2088116 lands as a separate PR), and (b) the leak fix is already in its own commit (4d04fbb, regression test included), I'd propose:

1. **Now**: I land 2088116 (cursor zero-row-skip fix + tests against main's signature) as the cherry-pickable first commit. Same for 093d692 (`getObject(Class)` base-class fallback).
2. **If you want the leak fix shipped independently**: I cherry-pick 4d04fbb's *intent* into a new PR against `main`, reshaping the fix to whatever the pre-unify code's leak surface actually is (likely a fix in `QueryResultArrowStream.toArrowStreamReader` to wrap the `ArrowStreamReader` so its close also closes the `RootAllocator` — cleaner than threading the allocator through `StreamingResultSet.of`).
3. **This PR stays as-is** for the leak fix: it lands as part of the unified refactor.

Lean toward not splitting the unify commit unless there's a strong reason. Want me to open the separate "main-targeted leak fix" PR per (2)?

mkaufmann:
Per the two review threads, split out the cherry-pickable fixes as their own PRs against `main`:

- **#185** — `fix: skip zero-row batches in ArrowStreamReaderCursor` (commit `2088116` from this branch, reshaped against `main`'s 2-arg cursor signature). Self-contained: only `ArrowStreamReaderCursor.java` + `ArrowStreamReaderCursorTest.java`.
- **#186** — `fix: provide default getObject(Class) fallback in QueryJDBCAccessor` (commit `093d692` from this branch, applies cleanly on `main`). Self-contained: only `QueryJDBCAccessor.java` + the bundled `StreamingResultSetMethodTest.getObjectWithClassUsesAccessorBaseFallback` regression test.

This PR (#175) keeps the same fixes as the first two commits — when #185 / #186 land, those commits will collapse to no-ops at rebase time.

For the remaining "should QueryResultArrowStream allocator-ownership move pre-unify too?" thread (#175 review): waiting on your call before I do that split. As I noted there, it's a non-trivial surgery on the unify commit and I'd rather get your sign-off before rewriting ~800 lines of refactor.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudResultSet.java
```diff
@@ -4,8 +4,511 @@
...
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import lombok.val;
+import org.apache.arrow.memory.BufferAllocator;
+import org.apache.arrow.vector.ipc.ArrowStreamReader;
 
-public interface DataCloudResultSet extends ResultSet {
-    String getQueryId();
+@Slf4j
+public class DataCloudResultSet implements ReadOnlyResultSet, ForwardOnlyResultSet, ResultSetWithPositionalGetters {
```
mkaufmann, line 49: **Blocking** — `DataCloudResultSet` was a `public interface` on `main` (just `extends ResultSet` plus `String getQueryId()`); after this PR it is a `public class` with private constructors. That is a binary- and source-incompatible change for anyone consuming the shaded fat JAR who:

- holds `DataCloudResultSet` as an interface reference and links against a recompiled-against-old-version classpath (`IncompatibleClassChangeError` at link time),
- writes `class MyRs implements DataCloudResultSet` for a delegate, decorator, or test double,
- reflects on `DataCloudResultSet.class.isInterface()`.

The in-tree migration is fine (`StreamingResultSet` → `DataCloudResultSet`), but external users have no fallback path because the *symbol* changed kind.

Prefer keeping `DataCloudResultSet` as an interface and renaming the new concrete class — `ArrowDataCloudResultSet` or a package-private `DataCloudResultSetImpl` would both keep the public symbol unchanged. If you accept the break instead, mark it as such (`refactor!:` subject + `BREAKING CHANGE:` footer so release-please picks it up — see the separate finding on release notes) and call it out in the body so users notice in the changelog rather than at deploy time.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

mkaufmann (commented):
## Review summary

12 findings: 1 Blocking, 8 Suggested, 3 Optional. The PR collapses the metadata path onto the streaming Arrow pipeline cleanly, and the close-order story for the new `Result` holder is right where it lands. The findings cluster around three themes:

1. **API surface.** The diff promotes `DataCloudResultSet` from an interface to a concrete class. That's a binary break for anyone implementing or proxying the type from the shaded JAR — flagged at `DataCloudResultSet.java:49` (Blocking). `StreamingResultSet`, `DataCloudMetadataResultSet`, `SimpleResultSet`, and `ColumnAccessor` all leave the public surface without a deprecation window. With release-please configured for Conventional Commits, the squash-merge subject `refactor:` will not surface a CHANGELOG entry — finding S7 covers this.

2. **JDBC-spec compliance regressions on metadata.** `MetadataSchemas.TYPE_INFO` declares `CASE_SENSITIVE` / `UNSIGNED_ATTRIBUTE` / `FIXED_PREC_SCALE` / `AUTO_INCREMENT` as `text(...)` columns while `HyperTypes.typeInfoRows()` populates them with `Boolean` values; the new `VarCharVectorSetter` `toString()`s them and `BaseIntVectorAccessor.getBoolean` is not overridden, so `getBoolean(...)` on these columns now throws where the old `SimpleResultSet` returned the raw `Boolean` (S1, S2). JDBC 4.2 Table B-6 lists `INTEGER → boolean` and `VARCHAR → boolean` (for "true"/"false"/"1"/"0") as recommended conversions; tools that read `NULLABLE` from `getColumns()` or boolean flags from `getTypeInfo()` will break.

3. **Allocator-hygiene gaps that survive the cleanup.** `cursor.close()` and `DataCloudResultSet.of`'s failure path use plain `try/finally` without `addSuppressed`, so if the leak detector trips on `allocator.close()` it masks the original exception (S3). `DataCloudStatement.executeQuery` / `getResultSet` calls `getQueryId()` between allocator construction and ownership transfer to `DataCloudResultSet.of` — if that gRPC call throws, the allocator escapes (S4).

**On the PR body itself:** The body describes a design that isn't in the diff. The "source-agnostic cursor" / `BatchLoader` / `AutoCloseable` ownership and the `StreamingResultSet.ofInMemory(root, owned, queryId, zone, cols)` factory don't exist in the cursor or in `DataCloudResultSet`. The metadata path actually round-trips through `ArrowStreamWriter` → IPC bytes → `ArrowStreamReader` (verified at `MetadataResultSets.java:88`), and the typeName-preservation mechanism rides on Arrow field metadata under a `datacloud-jdbc:type_name` key (`HyperTypeToArrow.toField` / `ArrowToHyperTypeMapper.toColumnMetadata`), not on a `cols` parameter. The body also references a non-existent `MetadataArrowBuilder` class (the factory is `MetadataResultSets`), names `StreamingResultSet` as the unified type when in fact `StreamingResultSet.java` is deleted and `DataCloudResultSet.java` is the new home, and the "Built on top of #`moritz/centralize-types-via-hypertype`" line is malformed (won't render as a PR ref) and stale (the dependency landed as #174). The first-line "I'm not happy with the PR and the approach" disclaimer would land in `git log` permanently if squashed as-is.

**Range-check tightening (`IntVectorSetter` / `SmallIntVectorSetter` / `TinyIntVectorSetter` now throw `IllegalArgumentException` on out-of-range Numbers — verified at `VectorPopulator.java:240,261,433`) is a real behavior change that the PR body's "Behavior changes worth calling out" section omits.** Worth a bullet so future debugging of `IllegalArgumentException` from prepared-statement parameter binding has a paper trail.

**Tests.** Behavior change "metadata int columns no longer accept `getBoolean`" is no longer pinned by a test — the old `assertThat(getBoolean("NULLABLE")).isFalse()` was replaced with `getInt("NULLABLE")`, and the test comment ("long→boolean coerces to false") is misleading because the code does not actually coerce — `BaseIntVectorAccessor` inherits `getBoolean` from `QueryJDBCAccessor`, which throws. A direct `assertThrows(SQLException.class, () -> getBoolean("ORDINAL_POSITION"))` would pin it.

Inline findings follow.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/VectorPopulator.java
```diff
@@ -118,36 +154,38 @@ abstract class BaseVectorSetter<T extends ValueVector, V> implements VectorValue
...
     VarCharVectorSetter() {
-        super(String.class);
+        super(Object.class); // accept String, Number, Boolean, byte[] — coerce to UTF-8 bytes
     }
 
     @Override
-    protected void setValueInternal(VarCharVector vector, String value) {
-        vector.setSafe(0, value.getBytes(StandardCharsets.UTF_8));
+    protected void setValueInternal(VarCharVector vector, int index, Object value) {
+        byte[] bytes =
```
mkaufmann: **Suggested** — Two related concerns on `VarCharVectorSetter`, which now widens from `String` to `Object` and `toString()`s anything that isn't `byte[]`:

*Type-safety regression.* The constructor comment says it accepts "String, Number, Boolean, byte[]", but the code doesn't enforce that — passing a `LocalDate`, a `Map`, or any foreign object writes its `toString()` into the vector instead of failing fast. The old `String`-only setter caught these at population time. This goes the opposite direction of the integer setters in this same file (lines 240, 261, 433 throw `IllegalArgumentException` on out-of-range `Number`). The asymmetry stings: integers refuse caller bugs, varchar swallows them. Either restrict to the documented set and throw `IllegalArgumentException` for anything else, or drop the comment and accept the permissive contract.

*JDBC spec regression on `getTypeInfo()`*. Because varchar is now permissive, `MetadataSchemas.TYPE_INFO` declaring `CASE_SENSITIVE` / `UNSIGNED_ATTRIBUTE` / `FIXED_PREC_SCALE` / `AUTO_INCREMENT` as `text(...)` (lines 49-53 on `main`, untouched here) combined with `HyperTypes.typeInfoRows()` populating them with raw `Boolean` values means the vector ends up storing `"true"`/`"false"` strings. `VarCharVectorAccessor.getBoolean()` does not handle the standard `"true"`/`"false"`/`"1"`/`"0"` coercion mandated by JDBC 4.2 Table B-6, so `rs.getBoolean("CASE_SENSITIVE")` against `getTypeInfo()` now throws `SQLFeatureNotSupportedException`. The old `SimpleResultSet` path returned the raw `Boolean`. BI tools that respect the spec (DBeaver, Tableau) break here.

Fix paths: (a) declare these specific columns as `bool(...)` in `MetadataSchemas` so `BitVectorSetter` handles `Boolean` natively — local and matches JDBC for these columns, or (b) implement spec-compliant `VARCHAR → boolean` coercion in `VarCharVectorAccessor.getBoolean()` — broader and fixes any future VARCHAR-with-boolean-content column at once.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
```diff
@@ -91,6 +103,12 @@ public boolean next() {
     @SneakyThrows
     @Override
     public void close() {
-        reader.close();
+        // Close the reader first: it releases the buffers accounted against the allocator, so the
+        // allocator's closing budget check passes. Reversing the order trips a leak detector.
+        try {
+            reader.close();
+        } finally {
```
mkaufmann: **Suggested** — `try { reader.close(); } finally { allocator.close(); }` with no `addSuppressed`. The javadoc above says reader-first ordering is load-bearing because the allocator's leak detector trips when buffers are still outstanding — meaning the most diagnostically interesting failure mode is exactly the one where reader.close() throws *and* allocator.close() then throws on top, and the second exception masks the first via Java's standard try/finally semantics.

Same pattern in `DataCloudResultSet.of`'s failure cleanup at line 96-103: `arrowStream.getReader().close()` is wrapped with `addSuppressed`, but the immediately-following `arrowStream.getAllocator().close()` is bare; if it throws (leak detector), it replaces the original construction `SQLException` rather than riding as suppressed.

```java
Throwable primary = null;
try { reader.close(); } catch (Throwable t) { primary = t; }
try { allocator.close(); } catch (Throwable t) {
    if (primary != null) primary.addSuppressed(t); else primary = t;
}
if (primary != null) throw primary;
```

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
```diff
@@ -197,7 +197,7 @@ public ResultSet executeQuery(String sql) throws SQLException {
             val arrowStream = SQLExceptionQueryResultIterator.createSqlExceptionArrowStreamReader(
                     iterator, includeCustomerDetail, iterator.getQueryStatus().getQueryId(), sql);
             resultSet =
-                    StreamingResultSet.of(arrowStream, iterator.getQueryStatus().getQueryId(), sessionZone);
+                    DataCloudResultSet.of(arrowStream, iterator.getQueryStatus().getQueryId(), sessionZone);
```
mkaufmann: **Suggested** — Once `arrowStream` is assigned at line 197-198, the 100 MB `RootAllocator` is alive and uncloseable except via `DataCloudResultSet.of` taking ownership. The very next statement evaluates `iterator.getQueryStatus().getQueryId()` *again* before the `of(...)` call. If that gRPC accessor throws (the catch on line 206 only catches `StatusRuntimeException`, and even that catch does not close `arrowStream`), the allocator escapes.

Same shape at line 437-441 in `getResultSet`: `adaptiveIterator.getQueryStatus().getQueryId()` is invoked again after `arrowStream` is constructed.

The PR's allocator-hygiene claim is "every code path closes its allocator". Easiest fix: hoist the queryId into a local before allocator construction:

```java
val qid = iterator.getQueryStatus().getQueryId();
val arrowStream = SQLExceptionQueryResultIterator.createSqlExceptionArrowStreamReader(
        iterator, includeCustomerDetail, qid, sql);
resultSet = DataCloudResultSet.of(arrowStream, qid, sessionZone);
```

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/MetadataResultSets.java
```diff
@@ -0,0 +1,148 @@
...
+     * unset (interpreted as Arrow null), which is almost always a caller bug. Today every caller
+     * goes through {@link MetadataSchemas} so the sizes match by construction; the precondition
+     * here makes a future caller bug surface at the boundary instead of in vector population.
+     */
+    public static DataCloudResultSet of(List<ColumnMetadata> columns, List<List<Object>> rows) throws SQLException {
+        validateRowArity(columns, rows);
+        byte[] ipcBytes = writeArrowStream(columns, rows);
+        // Allocator is handed to DataCloudResultSet along with the reader; the result set owns
+        // its lifecycle and closes it when close() is called.
+        RootAllocator allocator = new RootAllocator(Long.MAX_VALUE);
```
mkaufmann: **Suggested** — `RootAllocator allocator = new RootAllocator(Long.MAX_VALUE);` then `new ArrowStreamReader(...)` at line 66, then `DataCloudResultSet.of` at line 67. The `of(...)` factory has its own try/catch that closes both on construction failure, but it only fires once `of(...)` has been *entered* — if `new ArrowStreamReader(new ByteArrayInputStream(ipcBytes), allocator)` throws first, the allocator escapes. `QueryResultArrowStream.toArrowStreamReader` (the gRPC variant) has the same shape.

`ArrowStreamReader`'s constructor is benign in practice today, but the metadata path runs once per metadata query in production and the contract here advertises bulletproof allocator hygiene. Either wrap allocator+reader construction in a try/catch that closes the allocator on any throw before transfer, or factor out a small helper on `QueryResultArrowStream` that takes an `InputStream` + allocator size and returns a `Result` atomically; `MetadataResultSets.of` and `toArrowStreamReader` would both go through it.

Separately: `Long.MAX_VALUE` here while the query path is capped at `100 * 1024 * 1024` — the 100 MB cap exists for a reason, and a future caller hitting `getColumns(...)` against a customer with thousands of tables silently bypasses it. Reuse the same constant.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/DataCloudDatabaseMetadataTest.java
```diff
@@ -1082,9 +1082,14 @@ public void testTestTest() throws SQLException {
...
                 StringUtils.EMPTY, StringUtils.EMPTY, StringUtils.EMPTY, connection);
         while (columnResultSet.next()) {
+            // The metadata result set is Arrow-backed; TYPE_NAME carries "TEXT" (preserved from
+            // the JDBC-spec MetadataSchemas override), while TYPE_NAME's *value* is the HyperType's
+            // JDBC name ("VARCHAR" for varchar columns).
             assertThat(columnResultSet.getString("TYPE_NAME")).isEqualTo("VARCHAR");
             assertThat(columnResultSet.getInt("DATA_TYPE")).isEqualTo(12);
-            assertThat(columnResultSet.getBoolean("NULLABLE")).isFalse();
+            // NULLABLE is an INTEGER column. Arrow-backed getInt reports the nullability enum;
+            // 0 (columnNoNulls) for NOT NULL rows, which coerces to false via long→boolean.
```
mkaufmann: **Suggested** — The test comment says "`0` (columnNoNulls) for NOT NULL rows, which coerces to false via long→boolean", and the previous `getBoolean("NULLABLE")` assertion was replaced with `getInt("NULLABLE")`. But `BaseIntVectorAccessor` does not override `getBoolean()` — the call falls through to `QueryJDBCAccessor.getBoolean()` (line 52-54) which throws `SQLFeatureNotSupportedException`. The comment describes a coercion that does not happen in this codebase, and the behavior-change claim from the PR body ("`getBoolean` ... on an integer column throws `SQLException`") is no longer pinned by any direct assertion. Add:

```java
assertThrows(SQLException.class, () -> columnResultSet.getBoolean("ORDINAL_POSITION"));
```

next to the existing date/time `assertThrows` block — it pins the new contract and removes the misleading comment in one go. (See the separate finding on `MetadataSchemas` for the boolean-coercion gap that arguably *should* still work for spec compliance.)

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudResultSet.java
```diff
@@ -4,8 +4,511 @@
...
+     * throws (for example a {@link SQLException} wrapping an unsupported Arrow type), this
+     * method closes both before re-throwing so the 100 MB {@link
+     * org.apache.arrow.memory.RootAllocator} does not leak. Callers must not close either
+     * separately on success.
+     *
+     * @param arrowStream The Arrow stream + allocator pair, both owned by the result set.
+     * @param queryId The query identifier (may be {@code null} for synthesized result sets).
+     * @param sessionZone The session timezone used for timestamp conversions.
+     */
+    public static DataCloudResultSet of(QueryResultArrowStream.Result arrowStream, String queryId, ZoneId sessionZone)
```
mkaufmann, line 92: **Suggested** — Three deletions worth a CHANGELOG entry: `StreamingResultSet`, `DataCloudMetadataResultSet`, `SimpleResultSet`, `ColumnAccessor` were all `public` on `main`. Even if no in-tree code outside `:jdbc-core` referenced them, they were reachable from the shaded fat JAR. `StreamingResultSet.of(ArrowStreamReader, queryId)` and `of(ArrowStreamReader, queryId, ZoneId)` were the only public entry points for wrapping a foreign Arrow stream into a JDBC `ResultSet`. The new `DataCloudResultSet.of(QueryResultArrowStream.Result, ...)` requires constructing an internal `Result` holder, which is a tighter coupling than the old factories.

If there's no migration alias, please at least call this out in a `BREAKING CHANGE:` footer so release-please picks it up (the `refactor:` subject otherwise produces no CHANGELOG entry — see the release-notes finding). A thin `public static DataCloudResultSet of(ArrowStreamReader reader, BufferAllocator allocator, String queryId, ZoneId zone)` overload that internally builds a `Result` would also let external Arrow-stream wrappers continue to compile.

*Release-notes coverage.* Release-notes coverage. The repo uses release-please with `release-type: simple` and Conventional Commits. A squash-merge subject of `refactor: unify ResultSet implementations on Arrow-backed path` produces no CHANGELOG entry: refactor isn't on the release-please default sections, and at 0.x.y a `feat:` would be needed for a minor bump otherwise. Combined with the API surface deltas in this PR (interface→class on `DataCloudResultSet`, four removed public types, stricter accessor coercion on metadata int columns), the change will silently land in the next release and users discovering the break via `CHANGELOG.md` will have nothing to read.

Use `refactor!:` for the squash subject and add a `BREAKING CHANGE:` footer enumerating the specific deletions and behavior changes. release-please reads the `!` and the footer and produces a Breaking Changes section.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudResultSet.java
```diff
@@ -4,8 +4,511 @@
...
+
+    @Override
+    public boolean next() throws SQLException {
+        checkClosed();
+        return cursor.next();
+    }
+
+    @Override
+    public void close() throws SQLException {
+        if (!closed) {
```
mkaufmann: **Optional** — `close()` reads `if (!closed) { cursor.close(); closed = true; }`. If `cursor.close()` throws — e.g. the allocator's leak detector fires `IllegalStateException` — `closed` stays `false`. A defensive caller that catches and retries `close()` (try-with-resources won't, but JDBC connection pools and driver wrappers sometimes do) re-enters `cursor.close()`, which calls `allocator.close()` on an already-closed `RootAllocator`. Arrow throws on the second close.

Flip the order — set `closed = true` *before* delegating, or wrap in `try/finally` that sets it unconditionally. The caller still gets the cleanup exception; subsequent calls become no-ops, matching the standard JDK `AutoCloseable` pattern.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/VectorPopulator.java
```diff
@@ -260,13 +315,13 @@ class DecimalVectorSetter extends BaseVectorSetter<DecimalVector, BigDecimal> {
     }
 
     @Override
-    protected void setValueInternal(DecimalVector vector, BigDecimal value) {
-        vector.setSafe(0, value.unscaledValue().longValue());
+    protected void setValueInternal(DecimalVector vector, int index, BigDecimal value) {
```
mkaufmann, line 316: **Optional** — `DecimalVectorSetter.setValueInternal` does `vector.setSafe(index, value.unscaledValue().longValue())`. A `BigDecimal` whose unscaled exceeds `Long.MAX_VALUE` silently truncates. The PR added `IllegalArgumentException` range checks on `IntVectorSetter` (line 240), `SmallIntVectorSetter` (line 261), and `TinyIntVectorSetter` (line 433); decimal got the signature update but not the guard. The asymmetry is jarring now that this is the single populator path for both binding and metadata.

Not required (today's call sites probably stay in range), but worth a `bitLength() > 63 → IllegalArgumentException` check or a one-line javadoc explaining why decimals are exempt.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursorTest.java
```diff
@@ -37,12 +38,16 @@ class ArrowStreamReaderCursorTest {
...
     protected VectorSchemaRoot root;
 
+    @Mock
+    protected BufferAllocator allocator;
+
     @Test
     @SneakyThrows
-    void closesTheReader() {
-        val sut = new ArrowStreamReaderCursor(reader, ZoneId.systemDefault());
+    void closesReaderAndAllocator() {
```
mkaufmann, line 50: **Optional** — The new `closesReaderAndAllocator` test verifies both `reader.close()` and `allocator.close()` ran, but pins neither the order nor the throw-during-close contract. The cursor's javadoc explicitly says reader-first ordering exists "because reversing trips a leak detector", and the body uses `try/finally` so allocator.close runs even when reader.close throws. Both are exactly the properties most likely to regress.

A two-line tightening covers it:

```java
InOrder inOrder = Mockito.inOrder(reader, allocator);
inOrder.verify(reader).close();
inOrder.verify(allocator).close();
```

plus a second test where `reader.close()` is `doThrow(...)`-ed and the assertion is that `allocator.close()` still ran.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

mkaufmann:
Rebased on `origin/main` (now `7baf67d`), force-pushed. `:jdbc-core:test` passes.

Walked the 10 inline findings against the rebased tree — all still apply. PRs #185 (zero-row batch skip) and #186 (`getObject(Class)` identity fallback) landed on main since the review and dropped out of this PR's diff cleanly, but neither touches any of the lines a finding anchors to. The `IntegerVectorSetter` range-check tightening called out in the summary is still on this branch (`3bfd821`).

GitHub now flags the inline comments as outdated because of the SHA rewrite. The findings themselves are unchanged — text and line anchors are still correct.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudResultSet.java
mkaufmann: Going with **option B** — accepting the type-kind change and flagging it. Title flipped to `refactor!:` and PR body updated with a `BREAKING CHANGE:` footer enumerating the type-kind change, removed public types, and stricter int-column accessor semantics.

Reasoning: consumers of the fat JAR through the stock `java.sql.*` API (the supported integration path) recompile unchanged. Callers that depend on the concrete type would have had to adjust regardless given the metadata-path semantics change.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/VectorPopulator.java
mkaufmann: Fixed in 4c52910.

- VarCharVectorSetter tightened to `BaseVectorSetter<VarCharVector, String>` (super(String.class)) — non-String payloads now fail fast at the BaseVectorSetter type guard. The byte[] arm was dead (setBytes / setBinaryStream / setUnicodeStream / setAsciiStream all throw FEATURE_NOT_SUPPORTED in DataCloudPreparedStatement).
- TYPE_INFO's four boolean columns (CASE_SENSITIVE, UNSIGNED_ATTRIBUTE, FIXED_PREC_SCALE, AUTO_INCREMENT) now use a new `bool(...)` helper that produces `HyperType.bool(true)` / `Constants.BOOL` ColumnMetadata. Cross-checked against JDBC 4.2 (DatabaseMetaData.getTypeInfo Javadoc) and pgjdbc's PgDatabaseMetaData (which uses Oid.BOOL for the same four positions).
- HyperTypes.buildTypeInfoRow already passed Boolean values for these columns — no row-producer change needed once the schema declares them BOOL.
- Behavior change: `getObject` on the four columns now returns Boolean (per JDBC spec) instead of String. `rs.getBoolean(...)` starts working where it previously threw on the VARCHAR path; `rs.getString(...)` keeps returning lowercase "true"/"false" via `BooleanVectorAccessor.getString`.

Test coverage:
- `MetadataSchemasTest` gets three new tests pinning every TYPE_INFO column position by name / typeName / JdbcTypeId, mirroring the existing COLUMNS coverage.
- New `VarCharVectorSetterStrictTypeTest` (modeled on `IntegerVectorSetterRangeCheckTest`) regresses on Boolean / byte[] / Integer payloads so a future re-widening of the generic trips CI.
- `DataCloudDatabaseMetadataTest.testGetTypeInfo` now calls `rs.getBoolean(...)` on all four boolean columns end-to-end.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursor.java
mkaufmann: Fixed in 4029987.

- `ArrowStreamReaderCursor.close` now uses try-with-resources over `(allocator, reader)`. TWR closes in reverse declaration order — reader first (releases buffers, allocator's budget check passes), allocator second. If both throw, Java attaches the allocator's exception as suppressed onto the reader's instead of replacing it.
- `DataCloudResultSet.of` failure path: the bare `arrowStream.getAllocator().close()` is now wrapped in the same try/catch + `addSuppressed` pattern as the reader.close above it.

Pinned with a new Mockito test (`closeAttachesAllocatorErrorAsSuppressedWhenReaderCloseAlsoThrows`) that throws from both reader.close and allocator.close and asserts the reader exception is primary with the allocator exception attached as suppressed.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudStatement.java
mkaufmann: Fixed in 974efa7.

Hoisted `iterator.getQueryStatus().getQueryId()` (and `adaptiveIterator.getQueryStatus().getQueryId()` in `getResultSet`) into a local before `arrowStream` construction. Both call sites now compute the queryId once, hand it to `createSqlExceptionArrowStreamReader`, and reuse it for `DataCloudResultSet.of` — the only call between allocator construction and ownership transfer is `of(...)` itself, which has its own try/catch to clean up on construction failure.

No new test: the failure mode requires the gRPC accessor to throw between two adjacent statements, which is awkward to mock cleanly. The fix is code-shape — verified by reading and the existing test suite stays green.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/metadata/MetadataResultSets.java
mkaufmann: Fixed in dd484f2.

Both concerns addressed:

- **Construction window:** wrapped allocator + reader construction in try/catch at both sites (`MetadataResultSets.of` and `QueryResultArrowStream.toArrowStreamReader`). If the `ArrowStreamReader` constructor throws before `DataCloudResultSet.of` can take ownership, the allocator is closed on the way out with the close failure attached as suppressed.
- **Cap unification:** promoted `ROOT_ALLOCATOR_MB_FROM_V2` to a public `ROOT_ALLOCATOR_BUDGET_BYTES` constant on `QueryResultArrowStream` and use it from `MetadataResultSets`. Both paths now share the 100 MB cap, so a `getColumns(...)` against a tenant with thousands of tables trips a clean `OutOfMemoryException` from the allocator instead of letting the JVM OOM.

No new test — failure mode requires `ArrowStreamReader`'s constructor to throw, which doesn't happen with `ByteArrayInputStream` or the gRPC channel today. Pure code-shape fix; full `:jdbc-core:test` suite stays green.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/DataCloudDatabaseMetadataTest.java
mkaufmann: Fixed in 006cbda.

Took the spec-compliance route: added `getBoolean()` override to `BaseIntVectorAccessor` (one line, `getLong() != 0`) — JDBC 4.2 Table B-6 recommends `INTEGER → boolean` and pgjdbc / other major drivers do this. The previous behavior (inheriting `QueryJDBCAccessor`'s default that throws `SQLFeatureNotSupportedException`) broke spec-respecting clients reading metadata int columns as boolean.

Also:
- Restored `assertThat(columnResultSet.getBoolean("NULLABLE")).isFalse()` (the assertion deleted in the metadata-unify rebase) and added a positive case `assertThat(getBoolean("ORDINAL_POSITION")).isTrue()`.
- Corrected the comment that claimed the coercion already happened.

This pairs with [4c52910](https://github.com/forcedotcom/datacloud-jdbc/commit/4c52910) (TYPE_INFO boolean columns now declared as BIT). Every metadata int column that BI tools read as boolean now returns the spec-correct coercion. The `getObject(col, Boolean.class)` assertion at line 1138 still holds — that path goes through `QueryJDBCAccessor.getObject(Class)`'s `isInstance` fallback (added in #186), which is a separate surface and intentionally strict.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudResultSet.java
mkaufmann: PR title and body already cover this from thread #1 (refactor!: subject + BREAKING CHANGE: footer enumerating the deletions). Just refreshed the footer to reflect the spec-compliance work landed since:

- `getBoolean` on int columns no longer throws — coerces per JDBC 4.2 Table B-6 (commit 006cbda).
- `getTypeInfo()` boolean columns now declared BOOLEAN, not VARCHAR (commit 4c52910).

The `BREAKING CHANGE:` line release-please reads now accurately reflects what landed.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/DataCloudResultSet.java
mkaufmann: Fixed in f9633a1.

Flipped the flag before delegating, standard JDK AutoCloseable idempotence pattern:

```java
@Override
public void close() throws SQLException {
    if (closed) {
        return;
    }
    closed = true;
    cursor.close();
}
```

The caller still gets the cleanup exception on the first close; subsequent closes are no-ops.

Pinned with a new Mockito-based test (`closeIsIdempotentEvenIfFirstCloseThrows`) — spies a real `RootAllocator`, makes `close()` throw `IllegalStateException` (simulating the leak detector firing), asserts the first close rethrows, and verifies a retry doesn't re-enter the close path (`allocator.close` called exactly once across both attempts).

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/data/VectorPopulator.java
mkaufmann: Fixed in a9bb197.

Added the same `bitLength() > 63` guard the integer setters use for analogous narrowing:

```java
if (value.unscaledValue().bitLength() > 63) {
    throw new IllegalArgumentException("BigDecimal unscaled value " + value.unscaledValue()
            + " exceeds 64 bits — DECIMAL supports up to 18-digit unscaled longs");
}
vector.setSafe(index, value.unscaledValue().longValue());
```

Pinned with three tests mirroring `IntegerVectorSetterRangeCheckTest` style — accepts values within long range, rejects above, rejects below.

The separate scale-mismatch issue (`DecimalVectorSetter` ignores the vector's declared scale and writes whatever scale the caller's `BigDecimal` happens to have) is a real latent bug but out of scope for this finding — punt to a follow-up.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/ArrowStreamReaderCursorTest.java
mkaufmann: Fixed in 32ad2d9.

Tightened `closesReaderAndAllocator` with an explicit `InOrder` assertion:

```java
val order = inOrder(reader, allocator);
order.verify(reader).close();
order.verify(allocator).close();
```

The throw-during-close ordering is already pinned indirectly by the new `closeAttachesAllocatorErrorAsSuppressedWhenReaderCloseAlsoThrows` test (from [4029987](https://github.com/forcedotcom/datacloud-jdbc/commit/4029987)) — its `isSameAs(readerError).hasSuppressedException(allocatorError)` invariant only holds if reader closes first. So skipped option C (InOrder on both) — that would be belt-and-suspenders.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

praveen2450 (approved):

KaviarasuSakthivadivel (approved):

PR #177, opened 2026-04-30
feat: add support for using CDP auth token directly

reblem:
Adds support for using a CDP auth token directly instead of requiring a Core Salesforce token that must be exchanged for a CDP token. The CDP access token is accepted as a Bearer token via new `cdpToken` and `tenantUrl` connection properties, and validates the JWT eagerly.


jdbc/src/main/java/com/salesforce/datacloud/jdbc/DataCloudDatasource.java
```diff
@@ -161,6 +172,31 @@ private static DataCloudConnection createConnection(
                 dataspaceClient);
     }
 
+    private static DataCloudConnection createConnectionWithCdpToken(
```
KaviarasuSakthivadivel: This seems like a duplication code. Can we reuse the above createConnection method but tokenProvider needs to be modified so that it supports both the cases.

KaviarasuSakthivadivel (changes requested):

jdbc/src/main/java/com/salesforce/datacloud/jdbc/DataCloudDatasource.java
KaviarasuSakthivadivel: DONE.

KaviarasuSakthivadivel (approved):

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Optional;
+import java.util.Properties;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class DirectCdpTokenProcessor {
+    static final String CDP_TOKEN_KEY = "cdpToken";
+    static final String TENANT_URL_KEY = "tenantUrl";
+    static final String DATASPACE_KEY = "dataspace";
+    static final int FALLBACK_EXPIRES_IN_SECONDS = 3600;
```
mkaufmann: Why is this fallback required? The Bearer JWT should always have an exp section. Please check and otherwise remove the fallback and instead fail if exp is not ther

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+    private void validateToken() throws SQLException {
+        try {
+            DataCloudToken token = buildDataCloudToken();
+            token.getTenantId();
+            cachedDataCloudToken = token;
+        } catch (Exception ex) {
+            throw new SQLException(
+                    "Invalid CDP token: unable to parse JWT or extract tenant ID. " + ex.getMessage(), "28000", ex);
+        }
+    }
```
mkaufmann: This doesn't really seem to do validation, it only seems to check that the tenant id exists. Maybe `ensureTenantId()`?
Alternatively maybe merge this functionality into `buildDataCloudToken()`

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+        log.info("Lakehouse: {}", response);
+        return response;
+    }
+
+    private DataCloudToken buildDataCloudToken() throws SQLException {
+        DataCloudTokenResponse response = new DataCloudTokenResponse();
+        response.setToken(cdpToken);
+        response.setInstanceUrl(tenantUrl);
+        response.setTokenType("Bearer");
+        response.setExpiresIn(secondsUntilJwtExpiry(cdpToken));
```
mkaufmann: Should we fail here with a SQLException if it is already expired?

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public DataCloudToken getDataCloudToken() throws SQLException {
+        if (cachedDataCloudToken == null || !cachedDataCloudToken.isAlive()) {
+            cachedDataCloudToken = buildDataCloudToken();
+        }
+        return cachedDataCloudToken;
+    }
+
+    public String getLakehouse() throws SQLException {
```
mkaufmann: WOuld `getDatabaseName()` be a more descriptive name?

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+            if (exp == null || !exp.canConvertToLong()) {
+                return FALLBACK_EXPIRES_IN_SECONDS;
+            }
+            long remaining = exp.asLong() - Instant.now().getEpochSecond();
+            return remaining > 0 ? (int) Math.min(remaining, Integer.MAX_VALUE) : 0;
+        } catch (Exception ex) {
+            log.debug("Could not derive exp from CDP token, using default TTL", ex);
+            return FALLBACK_EXPIRES_IN_SECONDS;
+        }
+    }
```
mkaufmann, line 127: We should rather fail here with a SQLException

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+            String[] chunks = jwt.split("\\.", -1);
+            if (chunks.length < 2) {
+                return FALLBACK_EXPIRES_IN_SECONDS;
+            }
+            byte[] decoded = Base64.getUrlDecoder().decode(chunks[1]);
+            JsonNode payload = JSON.readTree(decoded);
+            JsonNode exp = payload.get("exp");
+            if (exp == null || !exp.canConvertToLong()) {
+                return FALLBACK_EXPIRES_IN_SECONDS;
+            }
```
mkaufmann, line 120: We should rather fail here with a SQLException

mkaufmann (approved):

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
KaviarasuSakthivadivel: Done — removed the fallback constant. expFromJwt() now throws SQLException when the JWT is malformed or has no numeric exp claim. Bearer JWTs are required to carry exp, so the fallback was hiding bugs.

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
KaviarasuSakthivadivel: Folded validateToken() into buildDataCloudToken() — token.getTenantId() runs at the end of construction, so a malformed JWT payload fails there rather than at first use.

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
KaviarasuSakthivadivel: Done — buildDataCloudToken() throws SQLException("CDP token has already expired") when exp <= now. Surfaces at ofDestructive, so users get a clear error before any gRPC call.

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
KaviarasuSakthivadivel: Renamed to getLakehouseName() to match the sibling DataCloudTokenProvider.getLakehouseName() on the OAuth path. Happy to rename both to getDatabaseName() if you'd prefer.

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
KaviarasuSakthivadivel: This is done as well.

jdbc-http/src/main/java/com/salesforce/datacloud/jdbc/auth/DirectCdpTokenProcessor.java
KaviarasuSakthivadivel: Yes, done., both branches now throw SQLException with descriptive messages.

KaviarasuSakthivadivel (approved):

mkaufmann (approved):

PR #178, opened 2026-04-30
feat: expose DataCloudStatement.getQueryStatus() with execution stats

KaviarasuSakthivadivel:
## Summary                                                                                                                                                                               
   
  Adds a non-blocking `DataCloudStatement#getQueryStatus()` accessor that returns                                                                                                          
  the latest observed wrapper `QueryStatus` (Trino-style — no extra RPCs). After
  iterating a `ResultSet` to completion, callers can read execution statistics                                                                                                             
  (wall-clock time, rows processed) from the status.
                                                                                                                                                                                           
  ## Design                                                                                                                                                                                
                                                                                                                                                                                           
  The status surface is split across two interfaces:                                                                                                                                       
                  
  - `protocol/RawQueryHandle` — proto-only view. Returns the gRPC                                                                                                                          
    `salesforce.cdp.hyperdb.v1.QueryStatus`. Implemented by the protocol-layer
    iterators and `AsyncQueryAccessHandle`. Used internally for query-id                                                                                                                   
    extraction, error reporting, and `cancel()`.                                                                                                                                           
  - `core/QueryHandle extends RawQueryHandle` — adds `getStatus()` returning the                                                                                                           
    public wrapper `com.salesforce.datacloud.query.v3.QueryStatus`. Created per                                                                                                            
    execution by `DataCloudStatement` as an anonymous class.                                                                                                                               
                                                                                                                                                                                           
  `DataCloudStatement` holds a single `QueryHandle queryHandle` field exposing                                                                                                             
  both views via the inheritance. The public `stmt.getQueryStatus()` delegates to                                                                                                          
  `queryHandle.getStatus()`.                                                                                                                                                               
                                                                                                                                                                                           
  ### Adaptive path                                                                                                                                                                        
  The iterator is the source of truth — its mutable proto field is updated as                                                                                                              
  the `ExecuteQuery` stream, chunk fetch, and `GetQueryInfo` polling progress.                                                                                                             
  The anonymous `QueryHandle` derives the wrapper from the iterator's current                                                                                                              
  proto on each read. No driver-side state.                                                                                                                                                
                                                                                                                                                                                           
  ### Async path                                                                                                                                                                           
  `AsyncQueryAccessHandle` captures only the initial proto (immutable). The                                                                                                                
  terminal status arrives later from `connection.waitFor(...)` inside                                                                                                                      
  `getResultSet()`. The driver writes that status into an                                                                                                                                  
  `AtomicReference<QueryStatus>` on `DataCloudStatement`; the anonymous                                                                                                                    
  `QueryHandle` reads from that reference. No `observeQueryStatus` mutator on                                                                                                              
  the protocol contract.                                                                                                                                                                   
                                                                                                                                                                                           
  ## Test plan                                                                                                                                                                             
                                                                                                                                                                                           
  - [x] Unit tests for both protocol iterators and `AsyncQueryAccessHandle`                                                                                                                
  - [x] Functional tests against local Hyper:
    - `getQueryStatusThrowsBeforeExecute`                                                                                                                                                  
    - `getQueryStatusAfterFullIterationAdaptivePathExposesExecutionStats`                                                                                                                  
    - `getQueryStatusAfterAsyncGetResultSetExposesExecutionStats`                                                                                                                          
  - [x] Example tests demonstrating both paths (`QueryStatusTest`)                                                                                                                         
  - [x] Full `:jdbc-core:test` green (1189 tests)


jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/QueryAccessHandle.java
```diff
@@ -4,12 +4,27 @@
...
+     */
+    QueryStatus getLatestWrapperStatus() throws SQLException;
+
+    /**
+     * Feeds an externally observed wrapper {@link QueryStatus} back into the handle. Handles that
+     * already track status via their own stream (e.g. the adaptive iterator) may ignore this call.
+     */
+    default void observeQueryStatus(QueryStatus status) {
+        // default: no-op
+    }
```
mkaufmann: This shouldn't be on the public QueryAccessHandle interface, this should be an internal implementation detail

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/QueryAccessHandle.java
```diff
@@ -4,12 +4,27 @@
...
  */
 public interface QueryAccessHandle {
-    QueryStatus getQueryStatus();
+    salesforce.cdp.hyperdb.v1.QueryStatus getQueryStatus();
+
+    /**
+     * Returns the latest observed query status as the public wrapper type.
+     * Implementations must never return {@code null} once the handle has been initialized.
+     */
+    QueryStatus getLatestWrapperStatus() throws SQLException;
```
mkaufmann: Most consumers should use `getQueryStatus` thous would keep that signature the same an return our own wrapped QueryStatus.

The new method should be something like
```
salesforce.cdp.hyperdb.v1.QueryStatus getQueryStatusProtoMessage();
```

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/AsyncQueryAccessHandle.java
```diff
@@ -6,22 +6,29 @@
...
 
 @Slf4j
-@AllArgsConstructor
 public class AsyncQueryAccessHandle implements QueryAccessHandle {
     @Getter
-    private final QueryStatus queryStatus;
+    private final salesforce.cdp.hyperdb.v1.QueryStatus queryStatus;
 
-    public static AsyncQueryAccessHandle of(HyperServiceGrpc.HyperServiceStub stub, QueryParam param) {
+    private volatile QueryStatus latestWrapperStatus;
```
mkaufmann: Why do we need two members? We should just keep one and construct the other on the fly. The current approach doesn't feel save with regards to potential divergence which wouldn't make sense here AFAIK

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/async/AsyncQueryResultIterator.java
```diff
@@ -191,4 +192,9 @@ public void close() {
     public QueryStatus getQueryStatus() {
         return queryStatus;
     }
+
+    @Override
+    public com.salesforce.datacloud.query.v3.QueryStatus getLatestWrapperStatus() throws SQLException {
+        return queryStatus == null ? null : com.salesforce.datacloud.query.v3.QueryStatus.of(queryStatus);
+    }
```
mkaufmann: As the protocol level classes operate on proto messages anyway what do you think about only returning the proto status message here? Does it make sense that this one implements QueryAccessHandle, is that relevant for the lower layer here?

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/QueryAccessHandle.java
KaviarasuSakthivadivel: yeah, that make sense. I have removed it.,

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/QueryAccessHandle.java
KaviarasuSakthivadivel: modified the interface name and kept getQueryStatus.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/async/AsyncQueryResultIterator.java
KaviarasuSakthivadivel: Yes, you are right, Only return the proto level classes make sense here. I have removed it.

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/protocol/AsyncQueryAccessHandle.java
KaviarasuSakthivadivel: removed it and have only one!

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/QueryHandle.java
```diff
@@ -0,0 +1,18 @@
...
+import java.sql.SQLException;
+
+/**
+ * Higher-level query handle held by {@link DataCloudStatement}. Extends {@link RawQueryHandle}
+ * (which exposes the gRPC proto status) with a wrapper-typed view used by the public
+ * {@link DataCloudStatement#getQueryStatus()} accessor.
+ */
+interface QueryHandle extends RawQueryHandle {
+    QueryStatus getStatus() throws SQLException;
+}
```
mkaufmann, line 18: It looks to me that the change would be simpler without the "extends RawQueryHandle". What do you think? You can drop if you disagree

mkaufmann (approved):

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/QueryHandle.java
KaviarasuSakthivadivel: That means, we need to maintain two status on DataCloudStatement, right? Especially for Async case.

j10t (approved):

PR #187, opened 2026-05-13
fix: don't return garbage values instead of null under arrow.enable_null_check_for_get=false

mkaufmann:
## Summary

When Iceberg is on the classpath it sets `-Darrow.enable_null_check_for_get=false` on the JVM for performance. With that flag off, Arrow's `VarCharVector`, `VarBinaryVector`, `FixedSizeBinaryVector`, and `TimeStamp*Vector` getters skip the validity check and return stale buffer bytes (often empty, but not guaranteed) for null rows instead of `null`. A SQL NULL surfaced as `""` / `0` / a sentinel timestamp, and `wasNull()` returned `false`. For example `SELECT NULL::timestamp` / `timestamptz` returned `1970-01-01T00:00:00Z` (or the `4714-12-31` sentinel for `getObject`); now - with the fix -returns `null`.

Behavior under the flag's default (`true`) is unchanged.

## Approach

Every accessor sources `wasNull` from `vector.isNull(int)` rather than from `holder.isSet` or a return-value sniff. `isNull` reads the validity buffer directly via `BitVectorHelper` — there's no per-row work for the flag to disable, so it is flag-proof.

This is a wider migration than the bug strictly requires. The flag-gating is currently only on VarChar/Binary/FixedSizeBinary/TimeStamp* in Arrow 17, but Arrow's contract ("skip the validity check on get") authorizes future minor versions to extend the gating to any vector type. Sourcing `wasNull` uniformly from `isNull` insulates us from any Arrow upgrade — including upgrades in consumer setups, where a tripwire test in this repository wouldn't fire.

`DataCloudArray.extractDataFromVector` had the same bug for VarChar/Binary/TimeStamp inner element types; it now consults `dataVector.isNull(absIndex)` before each `getObject` call.

## Test plan

`arrow.enable_null_check_for_get=false` is set in `buildSrc/src/main/kotlin/java-base-conventions.gradle.kts` `tasks.withType<Test>().configureEach`, so every test task across the build runs under the Iceberg scenario.

- Verified existing coverage catches the bug: with only the gradle flag and accessor sources reverted to `main`, **25 tests fail across 10 classes** — `JDBCReferenceTest`, `VarCharVectorAccessorTest`, `BinaryVectorAccessorTest`, `TimeStampVectorAccessorTest`, `DataCloudArrayTest`, `TimeZoneIntegrationTest`, plus metadata/connection paths asserting `expected: null but was: ""`.

### DIFF
--- a/jdbc-core/build.gradle.kts
+++ b/jdbc-core/build.gradle.kts
@@ -44,3 +44,11 @@ dependencies {
 tasks.named("compileJava") {
     dependsOn(":jdbc-grpc:compileJava")
 }
+
+// Iceberg sets -Darrow.enable_null_check_for_get=false on the JVM. With that flag off, Arrow's
+// VarCharVector/VarBinaryVector/FixedSizeBinaryVector .get(int) skip the validity check and return
+// stale buffer bytes for null rows instead of null. Run the test suite under that condition so
+// the existing null-handling assertions cover the Iceberg scenario.
+tasks.test {
+    systemProperty("arrow.enable_null_check_for_get", "false")
+}
--- a/jdbc-core/build.gradle.kts
+++ b/jdbc-core/build.gradle.kts
@@ -44,11 +44,3 @@ dependencies {
 tasks.named("compileJava") {
     dependsOn(":jdbc-grpc:compileJava")
 }
-
-// Iceberg sets -Darrow.enable_null_check_for_get=false on the JVM. With that flag off, Arrow's
-// VarCharVector/VarBinaryVector/FixedSizeBinaryVector .get(int) skip the validity check and return
-// stale buffer bytes for null rows instead of null. Run the test suite under that condition so
-// the existing null-handling assertions cover the Iceberg scenario.
-tasks.test {
-    systemProperty("arrow.enable_null_check_for_get", "false")
-}
--- a/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/VarCharVectorAccessor.java
+++ b/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/VarCharVectorAccessor.java
@@ -6,6 +6,7 @@ package com.salesforce.datacloud.jdbc.core.accessor.impl;
 
 import com.salesforce.datacloud.jdbc.core.accessor.QueryJDBCAccessor;
 import java.nio.charset.StandardCharsets;
+import java.util.function.IntPredicate;
 import java.util.function.IntSupplier;
 import org.apache.arrow.vector.LargeVarCharVector;
 import org.apache.arrow.vector.VarCharVector;
@@ -18,18 +19,20 @@ public class VarCharVectorAccessor extends QueryJDBCAccessor {
     }
 
     private final Getter getter;
+    private final IntPredicate nullChecker;
 
     public VarCharVectorAccessor(VarCharVector vector, IntSupplier currenRowSupplier) {
-        this(vector::get, currenRowSupplier);
+        this(vector::get, vector::isNull, currenRowSupplier);
     }
 
     public VarCharVectorAccessor(LargeVarCharVector vector, IntSupplier currenRowSupplier) {
-        this(vector::get, currenRowSupplier);
+        this(vector::get, vector::isNull, currenRowSupplier);
     }
 
-    VarCharVectorAccessor(Getter getter, IntSupplier currentRowSupplier) {
+    VarCharVectorAccessor(Getter getter, IntPredicate nullChecker, IntSupplier currentRowSupplier) {
         super(currentRowSupplier);
         this.getter = getter;
+        this.nullChecker = nullChecker;
     }
 
     @Override
@@ -39,9 +42,15 @@ public class VarCharVectorAccessor extends QueryJDBCAccessor {
 
     @Override
     public byte[] getBytes() {
-        final byte[] bytes = this.getter.get(getCurrentRow());
-        this.wasNull = bytes == null;
-        return this.getter.get(getCurrentRow());
+        // Arrow's vector.get(int) skips the isSet check when arrow.enable_null_check_for_get=false
+        // (e.g. set by Iceberg on the JVM), so a null entry returns an empty byte[] rather than null.
+        // Check the validity buffer explicitly via isNull(int).
+        final int row = getCurrentRow();
+        this.wasNull = nullChecker.test(row);
+        if (this.wasNull) {
+            return null;
+        }
+        return this.getter.get(row);
     }
 
     @Override
--- a/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/VarCharVectorAccessor.java
+++ b/jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/VarCharVectorAccessor.java
@@ -43,8 +43,8 @@ public class VarCharVectorAccessor extends QueryJDBCAccessor {
     @Override
     public byte[] getBytes() {
         // Arrow's vector.get(int) skips the isSet check when arrow.enable_null_check_for_get=false
-        // (e.g. set by Iceberg on the JVM), so a null entry returns an empty byte[] rather than null.
-        // Check the validity buffer explicitly via isNull(int).
+        // (e.g. set by Iceberg on the JVM), so a null entry returns stale buffer bytes (often empty,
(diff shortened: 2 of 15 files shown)


jdbc-core/build.gradle.kts
```diff
@@ -44,3 +44,11 @@ dependencies {
 tasks.named("compileJava") {
     dependsOn(":jdbc-grpc:compileJava")
 }
+
+// Iceberg sets -Darrow.enable_null_check_for_get=false on the JVM. With that flag off, Arrow's
+// VarCharVector/VarBinaryVector/FixedSizeBinaryVector .get(int) skip the validity check and return
+// stale buffer bytes for null rows instead of null. Run the test suite under that condition so
+// the existing null-handling assertions cover the Iceberg scenario.
+tasks.test {
+    systemProperty("arrow.enable_null_check_for_get", "false")
```
mkaufmann: **Suggested · finding 6** — Test flag is set only for `:jdbc-core`; other modules with vector tests run unprotected.

This sets `arrow.enable_null_check_for_get=false` only on `:jdbc-core`'s default `Test` task. `jdbc`, `integration-test`, `jdbc-http`, `spark-datasource-core`, and `spark-datasource` may exercise the same accessors via end-to-end paths and would not see the regression. The shared convention `buildSrc/src/main/kotlin/java-base-conventions.gradle.kts` already configures `tasks.withType<Test>()` (jvmArgs, useJUnitPlatform) — that is the right place. Two minor sub-points: `tasks.test { ... }` only targets the default `Test` task, so any future test task in `:jdbc-core` (integration test, separate junit suite) silently runs flag-on; and there is no in-source regression test pinning this issue, so if a future contributor migrates the build and accidentally drops the property, the entire coverage of this fix evaporates.

Move the `systemProperty("arrow.enable_null_check_for_get", "false")` into `tasks.withType<Test>().configureEach { ... }` in `java-base-conventions.gradle.kts`, keeping the explanatory comment.

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/VarCharVectorAccessor.java
```diff
@@ -39,9 +42,15 @@ public Class<?> getObjectClass() {
 
     @Override
     public byte[] getBytes() {
-        final byte[] bytes = this.getter.get(getCurrentRow());
-        this.wasNull = bytes == null;
-        return this.getter.get(getCurrentRow());
+        // Arrow's vector.get(int) skips the isSet check when arrow.enable_null_check_for_get=false
+        // (e.g. set by Iceberg on the JVM), so a null entry returns an empty byte[] rather than null.
```
mkaufmann: **Optional · finding 7** — Comment understates the failure mode; align with `BinaryVectorAccessor`.

This comment says "returns an empty `byte[]` rather than null". The matching comment in `BinaryVectorAccessor.getBytes` says "returns garbage rather than null" — accurate, since the bytes between offsets aren't guaranteed empty (cleared-buffer case happens to be empty, but a null row sandwiched between valid rows yields stale data). Aligning the two prevents future maintainers from reading the asymmetry as intentional.

Match the wording: "returns stale buffer bytes (often empty, but not guaranteed) rather than null".

---
_Generated by the `review-pr-tavern` skill — a human did not write this comment._

mkaufmann (commented):
**Review summary** — 7 findings (3 Blocking · 1 Suggested · 1 Optional summary-only; 2 inline). Findings 1–5 are summary-only because they reference files outside the diff hunks; 6 and 7 are inline.

---

### Blocking

**1. PR body claims `:jdbc-core:test` passes with the fix, but 10 tests still fail at PR HEAD.**

Running `./gradlew :jdbc-core:test --rerun-tasks` against `e3c304c` produces 10 failures: 3× `TimeStampVectorAccessorTest` (`testTimeStampMilliVector`, `testTimeStampMilliTZVector`, `testGetObjectWithUnsupportedType`), 3× `TimeZoneIntegrationTest` (`testInvalidTimezoneHandling`, `nullTimestampTZParameterStoresSqlNull`, `testGetObjectEqualsGetTimestamp` — `expected: null but was: 4714-12-31T23:52:58.000`), 2× `JDBCReferenceTest` (`select NULL::timestamp`, `select NULL::timestamptz`), 1× `DataCloudArrayTest::testGetArrayWithIndexPlusCountExceedingArraySize`, 1× `DataCloudConnectionFunctionalTest::testNetworkTimeoutPropagatesToServer`. Either the diff is incomplete or the test-plan claim is wrong. Findings 2 and 3 break this down.

**2. Timestamp accessors are still broken under flag-off — fix is incomplete.**
`jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/TimeStampVectorAccessor.java`

The PR body states that `TimeStamp` and `TimeStampTZ` "read `holder.isSet`, populated by `vector.get(int, holder)` which always honors validity regardless of the flag". The 6 failing TimeStamp tests above contradict that — null TIMESTAMP/TIMESTAMPTZ rows surface as `1970-01-01T00:00:00Z` or `4714-12-31T23:52:58.000` (sentinel). `TimeStampVectorGetter.createGetter` does use the holder pattern the body cites as safe, but for TimeStamp\* vectors that pattern is not actually flag-independent in Arrow 17. Apply the same `IntPredicate nullChecker` pattern to `TimeStampVectorAccessor` and `TimeStampTZVectorAccessor` — source `wasNull` from `vector.isNull(row)` rather than `holder.isSet`. Tighten the body's "all holders are validity-correct" claim to "all non-TimeStamp holders".

**3. `DataCloudArray` exposes Iceberg-mode garbage for null elements inside VarChar/VarBinary/FixedSizeBinary/TimeStamp arrays.**
`jdbc-core/src/main/java/com/salesforce/datacloud/jdbc/core/accessor/impl/DataCloudArray.java:61`

`extractDataFromVector` calls `dataVector.getObject(absIndex)` per element. For `VarCharVector` / `VarBinaryVector` / `FixedSizeBinaryVector` (and the TimeStamp\* variants per finding 2), `getObject` is also flag-gated under flag-off. `DataCloudArrayTest::testGetArrayWithIndexPlusCountExceedingArraySize` fails at PR HEAD with `expected: null but was: ""` — the bug philosophy of this PR ("source `wasNull` from the validity buffer") applies here too. Replace `dataVector.getObject(absIndex)` with a validity-aware lookup: if `dataVector.isNull(absIndex)` set the slot to null, otherwise call `getObject`.

### Suggested

**4. `*FromNulled*` accessor tests never assert `wasNull()` was set correctly.**
`jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/accessor/impl/VarCharVectorAccessorTest.java:65` (and the equivalents in `BinaryVectorAccessorTest`)

The fix's full contract is "`getBytes` / `getString` / `getObject` set `wasNull=true` on a null row even when `arrow.enable_null_check_for_get=false`". `testGetStringGetObjectAndGetObjectClassFromNulledVarCharVector` and the binary equivalents only assert that the returned value is null. `wasNull()` is set as a side effect of `getBytes()` in production code, but the tests never read it back — `QueryJDBCAccessorAssert.wasNull()` exists and is unused. A regression that fixed the return value but broke `wasNull` would slip past the unit tests. Append a `wasNull` assertion after each getter call in each `*FromNulled*` test — one line per test.

### Optional

**5. `nulledOutVector` reproduces a degenerate variant, not the realistic Iceberg failure mode.**
`jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/util/RootAllocatorTestExtension.java:150-153`

`nulledOutVector` calls `vector.clear()` then `setValueCount(original)`. `clear()` releases the buffers; `setValueCount` then allocates fresh zeroed buffers, so `vector.get(row)` reads `start=end=0` and returns an empty `byte[]` — enough to make the current `*FromNulled*` tests fail without the fix, but not the production failure mode. The realistic case is "row N is valid (writes real bytes), row N+1 is set null via `setNull(N+1)` (validity bit cleared, buffer bytes left over), getter returns N's stale bytes for N+1". Add a parameterised test that builds a vector with mixed valid/null entries via `setSafe` + `setNull` and walks the accessor across all rows asserting `null` / `wasNull()==true` for the null rows. Pins the realistic Iceberg scenario.

---

_Generated by the `review-pr-tavern` skill — a human did not write this comment._

KaviarasuSakthivadivel (approved):

praveen2450 (approved):

PR #201, opened 2026-09-23
fix(jdbc-core): decode oid columns as unsigned 32-bit integers

KaviarasuSakthivadivel:
Hyper's oid type (an unsigned 32-bit Postgres-style object identifier) arrives over Arrow as Int(32, signed=false). The driver ignored the sign flag in three places, so values >= 2^31 (e.g. 4294967295) came back through JDBC as negative ints instead of the correct unsigned value:

- ArrowToHyperTypeMapper classified the column as plain INTEGER metadata instead of BIGINT/oid.
- NumericGetter's UInt4Vector getter sign-extended the raw 32-bit pattern into a long (0xFFFFFFFF -> -1) instead of zero-extending it.
- BaseIntVectorAccessor routed UINT4 through getInt() (truncating) and was constructed with isUnsigned=false, corrupting getObject() and getString() as well.


jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/accessor/impl/BaseIntVectorAccessorTest.java
```diff
@@ -189,6 +191,13 @@ private List<Integer> getIntValues() {
                 Integer.MAX_VALUE);
     }
 
+    private List<Integer> getUInt4Values() {
+        // UInt4 (Hyper's oid) stores the full unsigned 32-bit range in a Java int's 32 raw bits.
+        // Values at or above 2^31 only round-trip correctly as an *unsigned* int, i.e. as a
+        // negative Java int (e.g. -1 is the raw bit pattern for the unsigned max 4294967295).
+        return ImmutableList.of(0, 1, Integer.MAX_VALUE, Integer.MIN_VALUE, -1);
```
mkaufmann: Lets extend jdbc-reference/src/main/resources/protocolvalues.json with the extreme values and rebuild the reference.json - that way we ensure we have same behavior as Postgres JDBC

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/protocol/data/HyperTypeArrowRoundtripTest.java
```diff
@@ -120,4 +122,25 @@ void timeTzAsymmetry_preservedOnInputButDroppedOnOutput() {
...
+    @Test
+    void oidAsymmetry_preservedOnInputButWidensOnOutput() {
+        // Not a bug — intended behavior (see HyperTypeToArrow's OID comment). Outbound, OID is
+        // encoded as a signed 64-bit Arrow int (large enough to hold the full unsigned 32-bit
+        // range without ambiguity) rather than the unsigned 32-bit Arrow int Hyper's server
+        // actually sends for a real oid column, so it decodes back as plain INT64, not OID.
+        HyperType original = HyperType.oid(true);
+        Field field = new Field("col", HyperTypeToArrow.toFieldType(original), null);
+        assertThat(ArrowToHyperTypeMapper.toHyperType(field)).isEqualTo(HyperType.int64(true));
+    }
```
mkaufmann: Why don't we use unsinge 32 bit for encoding in Arrow?

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/accessor/impl/BaseIntVectorAccessorTest.java
KaviarasuSakthivadivel: JDBCReferenceTest.getBaselineEntries() (line 101-104) currently hard-excludes every oid-typed column from the comparison outright — isTestable = false with the comment "This type behaves differently between Hyper and Postgres and is not worth it to test."
 Unfortunately we haven't been testing this oid. I will try to cover these case as well.

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/protocol/data/HyperTypeArrowRoundtripTest.java
KaviarasuSakthivadivel: Pushed a fix that gives OID its own case using Int(32, false), so it's now symmetric with the inbound mapping — HyperType.oid(...) round-trips through Arrow cleanly instead of widening to a plain int64.

jdbc-core/src/test/java/com/salesforce/datacloud/jdbc/core/accessor/impl/BaseIntVectorAccessorTest.java
KaviarasuSakthivadivel: modified it.

